PR #18, opened 2026-04-16
fix: suppress web notifications and enable microphone input

### LINKED ISSUE #14: enhancement: suppress web notification permission prompts — use native macOS notifications instead
## Summary

The WebUI can prompt for browser notification permission, which is irrelevant and confusing in a native macOS wrapper. Native apps use `UNUserNotificationCenter`, not web push.

## Current behavior

The WKWebView may surface a notification permission prompt. Even if granted, web notifications won't integrate with the macOS notification center properly.

## Proposed solution

Inject a `WKUserScript` at document start that intercepts `Notification.requestPermission` and always returns `"denied"`, suppressing the prompt entirely. The native notification feature (issue #8) is the right path for response alerts in the Mac app.

```swift
let script = WKUserScript(
    source: "Notification.requestPermission = function(cb) { if (cb) cb('denied'); return Promise.resolve('denied'); };",
    forMainFrameOnly: false
)
webView.configuration.userContentController.addUserScript(script)
```

## Related

- Issue #8 (feat: macOS notification when AI response completes) — the native replacement for this

### LINKED ISSUE #16: bug: microphone/audio input silently fails — WKWebView needs getUserMedia permission delegation
## Description

Voice/audio input does not work in the native macOS app. There is no macOS permission dialog asking the user to grant microphone access, and the feature fails silently.

## Root cause

The WebUI uses `navigator.mediaDevices.getUserMedia` for microphone access. In a standard browser (Chrome/Safari), this triggers a browser-level permission prompt. Inside `WKWebView`, this doesn't happen automatically — the app must:

1. Add `NSMicrophoneUsageDescription` to `Info.plist`
2. Implement `WKUIDelegate`'s `webView(_:requestMediaCapturePermissionFor:initiatedByFrame:type:decisionHandler:)` to forward the permission request to macOS
3. Optionally pre-request `AVAudioSession` access on first launch

Without this, `getUserMedia` is silently rejected and microphone input never works.

## Steps to reproduce

1. Open the native Hermes Agent app
2. Click the microphone/voice input button in the chat composer
3. No macOS permission dialog appears; audio does not work

## Expected behavior

macOS shows a permission dialog: "Hermes Agent would like to access your microphone." After granting, voice input works.

## Related

Similar to #14 (browser notification prompts) — these are all cases where browser-native permission APIs need to be bridged to native macOS equivalents via WKWebView delegate methods.

spinuplabs:
Resolves #14 — Suppress web notification permission prompt
Injects a WKUserScript at document start that overrides
Notification.requestPermission to always resolve as "denied",
preventing a browser-style permission dialog from appearing inside
the native wrapper. UNUserNotificationCenter is the intended path
for response alerts.

Resolves #16 — Microphone / voice input
Several changes work together to make voice input functional:

1. NSMicrophoneUsageDescription added to Info.plist (build.sh) so
   macOS will show the system permission dialog.

2. Implements WKUIDelegate's requestMediaCapturePermissionFor, which
   WKWebView requires before forwarding getUserMedia to the OS.
   Drives the full AVCaptureDevice authorization lifecycle —
   requesting the system dialog for .notDetermined, granting for
   .authorized, and denying for .denied/.restricted.

3. Adds requestMicrophonePermission() at app launch (AppDelegate) so
   the system dialog fires on first run before the user touches the
   mic button. If previously denied, shows a native NSAlert with an
   "Open System Settings" button linking directly to Privacy &
   Security → Microphone.

4. Injects a getUserMedia wrapper script that catches
   NotAllowedError/PermissionDeniedError, fires Escape to dismiss
   any hung spinner, and shows a toast directing the user to System
   Settings — auto-dismissed after 7 seconds.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -64,6 +64,7 @@ jobs:
           /usr/libexec/PlistBuddy -c "Add :LSMinimumSystemVersion string 12.0" "$APP_BUNDLE/Contents/Info.plist"
           /usr/libexec/PlistBuddy -c "Add :NSHighResolutionCapable bool true" "$APP_BUNDLE/Contents/Info.plist"
           /usr/libexec/PlistBuddy -c "Add :LSUIElement bool false" "$APP_BUNDLE/Contents/Info.plist"
+          /usr/libexec/PlistBuddy -c "Add :NSMicrophoneUsageDescription string 'Hermes Agent uses the microphone for voice input in the chat interface.'" "$APP_BUNDLE/Contents/Info.plist"
 
       - name: Create DMG
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## [v1.0.3] — 2026-04-16
+
+### Added
+- Microphone permission prompt at app launch — macOS shows the system dialog on first run before the user touches the mic button. If previously denied, a native alert appears with an "Open System Settings" button linking directly to Privacy & Security → Microphone. (PR #18, fixes #16, by @redsparklabs)
+- `requestMediaCapturePermissionFor` WKUIDelegate method — WKWebView now forwards microphone access requests through the macOS TCC authorization lifecycle before granting or denying `getUserMedia`. Without this, the browser `getUserMedia` call silently fails even when system permission is granted. (PR #18, fixes #16, by @redsparklabs)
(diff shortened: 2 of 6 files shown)


spinuplabs:
Still working on this one.  Don't review yet.

spinuplabs:
This gets speech recognition working, but it's slow.  
Ref: [hermes-webui Issue 590](https://github.com/nesquena/hermes-webui/issues/590)

nesquena-hermes:
@/tmp/pr18-review.txt

PR #19, opened 2026-04-16
fix: restore WebView focus so buttons and Cmd+K respond reliably

### LINKED ISSUE #13: bug: 'New conversation' button does nothing when a session is already open
## Bug description

The user was running hermes directly from his mac and complained of the new conversation button not working. 

The "New conversation" button has no effect when clicked after a conversation is already loaded. Clicking it multiple times does nothing.

## Steps to reproduce

1. Open the app
2. Select any previous conversation from the list — it loads in the chat area
3. Click the "New conversation" button (or press Cmd+K)
4. Nothing happens

## Expected behavior

Clicking "New conversation" should clear the chat area and start a fresh conversation.

## Actual behavior

The button click is silently ignored. The user has to find an alternative workaround.

## Notes

Likely a WKWebView focus/event-routing issue — click events on certain elements may not fire when the WebView doesn't have keyboard focus, or JavaScript event listeners may not be reached from the native wrapper. Worth checking whether a `window.focus()` call or a `WKScriptMessage` relay is needed before dispatching the button action.

spinuplabs:
Resolves #13

Summary
Clicks on buttons like "New conversation" had no effect after focus moved away from the WebView, and keyboard shortcuts like Cmd+K required an extra click to work.

Changes
Two changes work together:

HermesWebView subclass overrides acceptsFirstMouse to return true, so the click that refocuses the window also registers as a content click in JavaScript. Previously that first click was consumed entirely by focus restoration.

windowDidBecomeKey (NSWindowDelegate) calls webView.becomeFirstResponder() whenever the window becomes active, so keyboard shortcuts reach JavaScript immediately without requiring a manual click to restore focus first.

### DIFF
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -13,9 +13,15 @@ class BrowserWindow: NSWindow {
     }
 }
 
-class BrowserWindowController: NSWindowController, WKUIDelegate, WKNavigationDelegate {
+// Lets the first click on the WebView both focus it and register as a content
+// click simultaneously, fixing buttons that appear unresponsive after focus moves away.
+private class HermesWebView: WKWebView {
+    override func acceptsFirstMouse(for event: NSEvent?) -> Bool { true }
+}
 
-    private var webView: WKWebView!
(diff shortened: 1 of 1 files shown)


nesquena-hermes:
@/tmp/pr19-review.txt

PR #30, opened 2026-04-19
feat: v1.1.0 — launch at login, zoom shortcuts, status indicator, ATS fix, dark mode FOUC, error screen repo link

### LINKED ISSUE #3: feat: opt-in launch at login (pre-checked in setup)
## Overview

Add a "Launch at Login" option so Hermes is always ready in the menu bar. Should be opt-in but pre-checked during first-run setup, matching the pattern Claude.app, Notion, and most menu bar tools use.

## Why

A menu bar app that doesn't auto-start loses most of its value — users have to remember to open it. Pre-checking the option during setup is the correct default for a tool people will use daily. Opt-out is always available.

## Behaviour spec

- On first launch, after the user completes Preferences setup, show a one-time prompt:
  > "Launch Hermes at login so it's always ready in your menu bar?"
  > [✓ Launch at Login] [Not Now]
- Default: checked
- Setting persists in UserDefaults key `launchAtLogin`
- Also exposed as a checkbox in the Preferences window under a new "General" section
- Toggling the checkbox immediately registers/unregisters the login item — no save button needed for this field

## Implementation

Use `SMAppService` (macOS 13+) with a fallback to `SMLoginItemSetEnabled` for macOS 12:

```swift
import ServiceManagement

func setLaunchAtLogin(_ enabled: Bool) {
    if #available(macOS 13.0, *) {
        do {
            if enabled {
                try SMAppService.mainApp.register()
            } else {
                try SMAppService.mainApp.unregister()
            }
        } catch {
            print("SMAppService error: \(error)")
        }
    } else {
        SMLoginItemSetEnabled("ai.get-hermes.HermesAgent" as CFString, enabled)
    }
}
```

## Acceptance criteria

- [ ] First-launch prompt appears once, pre-checked
- [ ] "Launch at Login" checkbox in Preferences → General section
- [ ] Toggling immediately takes effect (verify by logging out and back in)
- [ ] Works on macOS 12 (SMLoginItemSetEnabled) and macOS 13+ (SMAppService)
- [ ] UserDefaults key `launchAtLogin` reflects current state

### LINKED ISSUE #23: bug: white flash on startup in dark mode — theme loads after initial WKWebView render
## Bug description
On launch in dark mode, the app briefly shows a white window before the dark theme loads — a classic FOUC (Flash of Unstyled Content) caused by the WKWebView rendering before the theme preference is applied.

*Reported by a single user — flagged for investigation.*

## Steps to reproduce
1. Enable dark mode (system or in-app)
2. Launch the Hermes Mac app
3. Observe: brief white flash before the UI renders in dark mode

## Expected behavior
Window background should be dark immediately on launch — no flash.

## Reference
Epicenter (another WKWebView-based app) solved the same issue in v7.11.0:
https://github.com/EpicenterHQ/epicenter/issues/1146

## Suggested fix
In the WKWebView setup, set the background color before content loads:

```swift
// Option 1: underPageBackgroundColor (macOS 12+)
webView.underPageBackgroundColor = NSColor(named: "AppBackground") // match dark bg

// Option 2: documentStart userScript
let script = WKUserScript(
    source: "document.documentElement.style.background = '#111111'",
    forMainFrameOnly: true
)
webView.configuration.userContentController.addUserScript(script)
```

Either approach ensures the background matches the dark theme before the first paint.

### LINKED ISSUE #24: feat: UI zoom/scale controls for high-PPI displays — pinch-to-zoom or Cmd+/- shortcuts
## Summary
On Mac with high-PPI (Retina) displays, there is no way to zoom or scale the UI. The WKWebView renders at a fixed scale with no mechanism to increase text/UI size for users who want a larger or denser view.

## Request
Add a way to zoom or scale the UI — either:

1. **Pinch-to-zoom** — enable WKWebView's built-in `allowsMagnification` so users can pinch to scale on trackpad
2. **Fixed zoom level setting** — a preference (e.g. 90% / 100% / 125% / 150%) that persists across sessions, set via `webView.pageZoom`
3. **Keyboard shortcuts** — Cmd+Plus / Cmd+Minus / Cmd+0 to zoom in/out/reset (standard macOS convention)

Option 3 (keyboard shortcuts) is the lowest-effort starting point and matches native macOS app expectations. Options 1 and 2 give more persistent control.

## Current behavior
No zoom or scale controls exist. On high-PPI displays the UI may feel too small with no way to adjust.

## Environment
- Mac (reported via Cloudflare Tunnel + local access)
- High PPI display

## Implementation notes
```swift
// Allow pinch-to-zoom (option 1)
webView.allowsMagnification = true

// Programmatic zoom (options 2 & 3)
webView.pageZoom = 1.25  // 125%

// Keyboard shortcuts — add to keyDown handler or menu
// Cmd+= → webView.pageZoom *= 1.1
// Cmd+- → webView.pageZoom /= 1.1
// Cmd+0 → webView.pageZoom = 1.0
```

Reported by @uzairansar on X.com.

### LINKED ISSUE #25: Non-localhost target URLs silently fail — ATS blocking + server binding (two compounding bugs)
## Non-localhost target URLs silently fail — two compounding bugs

**Summary:** When a user configures a non-`localhost` target URL in Preferences (e.g. a Tailscale IP like `http://100.76.126.48:8787`), the Mac app silently fails with "Cannot connect to Hermes." There are two independent issues compounding each other, and the error dialog gives no hint about either one.

---

### 1. App Transport Security (ATS) blocks plain HTTP to non-localhost addresses

WKWebView enforces ATS by default. `http://localhost` is automatically exempt, but any other `http://` URL — including Tailscale IPs (`100.x.x.x`), LAN IPs, or hostnames — is blocked at the networking layer before the connection is even attempted. The generated `Info.plist` (built in `build.sh:35–63`) contains no `NSAppTransportSecurity` key, so there's no exemption.

This means: even if the backend server is reachable on the network, WKWebView will refuse to load it. The load fails silently.

**Fix:** Add to the `Info.plist` heredoc in `build.sh`:
```xml
<key>NSAppTransportSecurity</key>
<dict>
    <key>NSAllowsArbitraryLoadsInWebContent</key>
    <true/>
</dict>
```

A more targeted alternative is to allow only `http` scheme loads in web content, which is the right scope for WKWebView while not opening the whole app to arbitrary ATS bypasses.

### 2. The Hermes server only binds to `127.0.0.1` by default

`bootstrap.py` reads `HERMES_WEBUI_HOST` from the environment (default: `127.0.0.1`). On a remote server, the process only listens on the loopback interface — so connections from any other address, including Tailscale (`100.x.x.x`), are refused at the OS level before even reaching the app.

**Server-side fix** (user has to do this manually):
```bash
# Add to .env in hermes-webui-public/:
HERMES_WEBUI_HOST=0.0.0.0
```
Or:
```bash
bash start.sh --host 0.0.0.0
```

Note: if no access password is set, the server will print a `[!!] WARNING` — binding to `0.0.0.0` without a password is intentionally discouraged.

### 3. Silent failure — no diagnostic information surfaced

`BrowserWindowController` is set as the `WKNavigationDelegate` but implements none of the failure callbacks:

- `webView(_:didFailNavigation:withError:)` — fires on mid-load errors
- `webView(_:didFailProvisionalNavigation:withError:)` — fires when the initial request fails (connection refused, ATS block, DNS failure)

Both ATS violations and connection-refused errors land in `didFailProvisionalNavigation`. Without implementing it, the app has no way to distinguish "server unreachable" from "ATS blocked the request" from "wrong port." The current "Cannot connect" dialog is shown from a health-check timeout, not from the actual WKWebView error — so the underlying cause is never visible.

---

## Impact

Any user running Hermes on a remote server (VPS, NAS, homelab) and trying to connect via Tailscale, a LAN IP, or a hostname will hit this. It's a common and completely reasonable deployment pattern, especially for users who don't want to set up an SSH tunnel manually.

---

## Workaround for the current version

There are two ways to work around this today. **Option B is strongly recommended** because it sidesteps both issues simultaneously.

### Option A — bind server to all interfaces + set a password

On the server, add to `.env` in the `hermes-webui-public/` directory:
```
HERMES_WEBUI_HOST=0.0.0.0
HERMES_WEBUI_PASSWORD=your-password-here
```
Then restart: `bash start.sh`

**Important:** This addresses the binding issue but **ATS will still block the connection** in the current Mac app. Option A alone is not sufficient until the ATS fix ships. You'd need to combine it with Option B.

### Option B — SSH tunnel (recommended, works today, no app changes needed)

Set up a tunnel from your Mac to the server:
```bash
ssh -N -L 8787:127.0.0.1:8787 user@100.76.126.48
```
(Replace `user@100.76.126.48` with your actual SSH user and Tailscale/server IP.)

Then in Hermes Agent Preferences, set the target URL back to:
```
http://localhost:8787
```

`localhost` is ATS-exempt, so the Mac app will load it fine. The SSH tunnel handles forwarding to the actual server. This works without any server-side changes.

---

### Priority 1 — ATS key in Info.plist (one-line fix, immediate)

In `build.sh`, inside the Info.plist heredoc:
```xml
<key>NSAppTransportSecurity</key>
<dict>
    <key>NSAllowsArbitraryLoadsInWebContent</key>
    <true/>
</dict>
```

This unblocks Tailscale IPs, LAN IPs, and hostnames without requiring HTTPS everywhere.

### Priority 2 — Implement navigation failure callbacks

In `BrowserWindowController.swift`, conform to WKNavigationDelegate and implement:

```swift
func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
    // Surface the actual error — ATS violations, connection refused, etc.
    showConnectionError(underlyingError: error)
}

func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
    showConnectionError(underlyingError: error)
}
```

The error message in the dialog should include:
- The underlying `NSURLError` description
- A hint if it looks like an ATS violation (`NSURLErrorAppTransportSecurityRequiresSecureConnection`)
- A hint if it looks like connection refused (`NSURLErrorCannotConnectToHost`)

### Priority 3 — Smarter error message for non-localhost HTTP

If the configured URL is `http://` and the host is not `localhost`/`127.0.0.1`, the "Cannot connect" dialog could proactively explain:
1. Make sure the server is running with `HERMES_WEBUI_HOST=0.0.0.0`
2. Or use an SSH tunnel and point the app at `http://localhost:8787`

This doesn't require detecting the actual failure cause — the URL shape alone is enough to show useful guidance.

---

## How to confirm it's ATS (diagnostic step)

Run the app from Terminal to see log output:
```bash
/Applications/Hermes\ Agent.app/Contents/MacOS/HermesAgent
```

Then open Console.app, filter by process `HermesAgent` — ATS violations appear as `CFNetwork` / `NSURLErrorDomain` messages with domain `-1022` or mention of `App Transport Security`.

Alternatively: set the target URL to `https://example.com` in Preferences. If that loads, ATS is definitely the cause.

---

## Environment

- macOS app: all current versions (no ATS key present in any build)
- Hermes WebUI: all versions (default binding is `127.0.0.1`)
- Affects: any non-localhost target URL over plain HTTP

### LINKED ISSUE #27: ux: error screen should name the backend repo URL, not just the path
## Problem

When the app launches without a backend running, the error screen says:

> Make sure hermes-webui is running on the target URL. Start it with:
> `cd ~/hermes-webui-public && bash start.sh`

The path `~/hermes-webui-public` is named, but the *repository* (`nesquena/hermes-webui`) is not. A first-time user who doesn't have the backend yet has no way to find it — the path alone is un-Googleable and doesn't appear anywhere in the DMG.

## User quote (beta tester)

> "The error screen names the path (`~/hermes-webui-public`) but never names the repo (`nesquena/hermes-webui`). Those are two different strings, and without the second one, you can't Google your way out."
> Total fix time once they knew what was wrong: ~90 seconds. Time spent figuring out what was wrong: significantly more.

## Fix

Add one line to the error screen below the `start.sh` instruction:

> Don't have it yet? **github.com/nesquena/hermes-webui**

This alone likely eliminates 80% of "it doesn't work" reports from new users.

## Suggested copy (error screen)

```
Cannot connect to Hermes

Make sure hermes-webui is running at http://localhost:8787

Start it with:
  cd ~/hermes-webui-public && bash start.sh

Don't have it yet? → github.com/nesquena/hermes-webui
```

## Also consider

- A "Get Hermes Web UI" button that opens `https://github.com/nesquena/hermes-webui` in the browser
- Or a "Set Up Backend…" button that opens a sheet with clone + start instructions

### LINKED ISSUE #29: ux: no connection status indicator in main UI — unclear when backend drops
## Problem

Once connected, there's nothing in the main UI showing which backend URL is active or that the connection is healthy. The only feedback is the full "Cannot connect" error screen when the backend is completely down.

A user switching between local and remote Hermes instances (via Preferences → URL) has no confirmation that the switch took effect.

## Suggested fix

A small status pill in the toolbar or title bar area:

- 🟢 `localhost:8787` — connected
- 🔴 `localhost:8787` — disconnected (tap to retry)

Could be as simple as a colored dot next to the URL in the window title, or a toolbar icon that reflects health check status. The backend already exposes `GET /health` which returns `{status: ok}`.

nesquena-hermes:
## v1.1.0 sprint — 6 quality-of-life fixes + 1 new feature

This PR addresses the most user-facing pain points from the issue backlog. All changes are self-contained — no new dependencies, no Info.plist key changes that require Entitlements.plist parity, no architectural changes.

---

### Fixed

**#27 — Error screen now links to the backend repo**
First-time users who hit "Cannot connect" had no way to find the backend. The path `~/hermes-webui-public` is un-Googleable. Added a clickable `github.com/nesquena/hermes-webui` link (direct mode only — SSH users already know the repo).

**#23 — Dark mode white flash on startup (FOUC)**
WKWebView rendered white before the dark theme loaded. Two-layer fix: `underPageBackgroundColor = .windowBackgroundColor` (macOS 12+) sets the native backing color before the first paint; a `documentStart` userScript sets the HTML `background` for the brief window before CSS loads.

**#25 — Non-localhost URLs silently failing (ATS)**
`http://` to Tailscale IPs, LAN addresses, or custom hostnames was blocked by WKWebView's App Transport Security with no visible error. Added `NSAllowsArbitraryLoadsInWebContent = true` to Info.plist in both `build.sh` and the CI workflow (parity maintained). `NSAllowsArbitraryLoads` stays `false` so URLSession (used for preflight HTTP checks) keeps its ATS restrictions.

---

### Added

**#24 — Cmd+/Cmd− zoom keyboard shortcuts**
Pinch-to-zoom was already enabled (`allowsMagnification = true`). This adds a View menu with Zoom In (`⌘+`), Zoom Out (`⌘-`), and Actual Size (`⌘0`) that call `webView.magnification`. The implementation routes through a `webViewForZoom` accessor on `BrowserWindowController` to avoid making the private `webView` ivar public.

**#29 — Connection status indicator in window title**
Direct mode showed no feedback once connected — no indication of which backend URL was active or whether it was healthy. The window title now shows `Hermes Agent  ● localhost:8787` (filled dot = healthy, hollow = offline). A Timer polls `/health` every 30s and updates the title only when status changes. The timer is cancelled in `windowWillClose`.

**#3 — Launch at login**
Preferences → APP now includes a "Launch at login" checkbox. Implementation:
- macOS 13+: `SMAppService.mainApp.register()` / `await .unregister()`
- macOS 12: checkbox renders disabled with an inline "Requires macOS 13" note
- `.requiresApproval` state surfaces a nudge to System Settings → General → Login Items
- No entitlement or Info.plist change needed (SMAppService.mainApp works for non-sandboxed Developer ID apps)
- No UserDefaults mirroring — `SMAppService.mainApp.status` is the single source of truth

---

### Build note for Mac agent review

Swift cannot run on the Linux server. **Before merging, please ask Nathan to run on the Mac agent:**

```
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-1-0 && git pull

swift package resolve
swift build -c release
swift test
bash build.sh

# Verify Info.plist has ATS keys:
/usr/libexec/PlistBuddy -c "Print :NSAppTransportSecurity" "Hermes Agent.app/Contents/Info.plist"

# Verify Sparkle keys still present:
```

Expected: `swift build` succeeds, `swift test` passes, `bash build.sh` produces `Hermes Agent.app`.

Closes #27
Closes #23
Closes #25
Closes #24
Closes #29
Closes #3

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -98,6 +98,11 @@ jobs:
           /usr/libexec/PlistBuddy -c "Add :LSUIElement bool false"                          "$APP_BUNDLE/Contents/Info.plist"
           /usr/libexec/PlistBuddy -c "Add :NSMicrophoneUsageDescription string 'Hermes Agent uses the microphone for voice input in the chat interface.'" "$APP_BUNDLE/Contents/Info.plist"
           /usr/libexec/PlistBuddy -c "Add :NSUserNotificationUsageDescription string 'Hermes Agent notifies you when an AI response is ready while the window is in the background.'" "$APP_BUNDLE/Contents/Info.plist"
+          # ATS: allow plain http:// in WKWebView for non-localhost targets (Tailscale, LAN, custom hostnames)
+          # NSAllowsArbitraryLoads stays false so URLSession (preflightHTTP) keeps normal ATS restrictions.
+          /usr/libexec/PlistBuddy -c "Add :NSAppTransportSecurity dict" "$APP_BUNDLE/Contents/Info.plist"
+          /usr/libexec/PlistBuddy -c "Add :NSAppTransportSecurity:NSAllowsArbitraryLoads bool false" "$APP_BUNDLE/Contents/Info.plist"
+          /usr/libexec/PlistBuddy -c "Add :NSAppTransportSecurity:NSAllowsArbitraryLoadsInWebContent bool true" "$APP_BUNDLE/Contents/Info.plist"
           # Sparkle keys
           /usr/libexec/PlistBuddy -c "Add :SUPublicEDKey string 'daAlTqdBbYPSCDjS9IfTCJOFDo1jqtjMRZluhtAbKMY='" "$APP_BUNDLE/Contents/Info.plist"
           /usr/libexec/PlistBuddy -c "Add :SUFeedURL string 'https://hermes-webui.github.io/hermes-swift-mac/appcast.xml'" "$APP_BUNDLE/Contents/Info.plist"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [v1.1.0] — 2026-04-19
+
+### Added
+- **Launch at login** — Preferences now includes a "Launch at login" checkbox (macOS 13+, SMAppService). Correctly handles `.requiresApproval` state and shows a System Settings nudge. Disabled with an inline note on macOS 12. (fixes #3)
+- **View zoom — Cmd+/Cmd− keyboard shortcuts** — new View menu with Zoom In, Zoom Out, and Actual Size. Pinch-to-zoom was already enabled; this adds keyboard control. (fixes #24)
+- **Connection status in window title** — direct-mode window title now shows the active backend host and a live indicator (● connected / ○ offline). A 30-second health poll detects when the backend goes away without a full page reload. (fixes #29)
+
+### Fixed
+- **Non-localhost URLs silently failing** — plain `http://` to Tailscale IPs, LAN addresses, and custom hostnames was blocked by App Transport Security in WKWebView. Added `NSAllowsArbitraryLoadsInWebContent` to Info.plist (both `build.sh` and CI workflow). URLSession (used for connection preflight) keeps default ATS restrictions. (fixes #25)
+- **Dark mode white flash on startup** — WKWebView rendered white before the dark theme loaded. Set `underPageBackgroundColor` to `.windowBackgroundColor` (macOS 12+) and added a `documentStart` userScript to set the HTML background before first paint. (fixes #23)
+- **Error screen didn't name the backend repo** — first-time users who hit "Cannot connect" had no way to find the backend. Added a clickable "github.com/nesquena/hermes-webui" link to the error screen in direct mode. (fixes #27)
+
 ## [v1.0.9] — 2026-04-16
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -248,11 +248,38 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         windowMenu.addItem(
             withTitle: "Zoom", action: #selector(NSWindow.zoom(_:)), keyEquivalent: "")
 
+        let viewMenuItem = NSMenuItem()
+        menuBar.addItem(viewMenuItem)
+        let viewMenu = NSMenu(title: "View")
+        viewMenuItem.submenu = viewMenu
+        viewMenu.addItem(
+            withTitle: "Zoom In", action: #selector(zoomIn), keyEquivalent: "+")
+        viewMenu.addItem(
+            withTitle: "Zoom Out", action: #selector(zoomOut), keyEquivalent: "-")
+        viewMenu.addItem(
+            withTitle: "Actual Size", action: #selector(zoomReset), keyEquivalent: "0")
+
         NSApp.mainMenu = menuBar
     }
 
     @objc func checkForUpdates() {
         updaterController.checkForUpdates(nil)
+    }
+
+    // MARK: - View zoom (fix #24)
+
+    @objc func zoomIn() {
+        guard let webView = browserWindow?.webViewForZoom else { return }
+        webView.magnification = min(webView.magnification + 0.1, 3.0)
+    }
+
+    @objc func zoomOut() {
+        guard let webView = browserWindow?.webViewForZoom else { return }
+        webView.magnification = max(webView.magnification - 0.1, 0.5)
+    }
+
+    @objc func zoomReset() {
+        browserWindow?.webViewForZoom?.magnification = 1.0
     }
 
     @objc func openPreferences() {
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -25,6 +25,11 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
 
     private var webView: HermesWebView!
     private var statusBar: NSView!
+
+    /// Exposes the WKWebView for zoom operations called from AppDelegate menu actions.
+    /// Return type is WKWebView (not the private HermesWebView subclass) so Swift's
+    /// access-level rules are satisfied — callers only need .magnification anyway.
+    var webViewForZoom: WKWebView? { webView }
     private var separator: NSView!
     private var statusDot: NSView!
     private var statusLabel: NSTextField!
@@ -34,6 +39,11 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
     private let connectionMode: String
     var onReconnect: (() -> Void)?
     var onNavigationFailed: (() -> Void)?
+
+    // Health check timer for direct mode — polls /health every 30s and
+    // reflects status in the window title (fix #29).
+    private var healthTimer: Timer?
+    private var isHealthy: Bool = true
 
     init(urlString: String, title: String, connectionMode: String = "direct") {
         self.urlString = urlString
@@ -48,6 +58,9 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         )
         window.title = title
         window.center()
+        // Fix #23: set native window background to dark before content loads,
+        // so there's no white frame visible while WKWebView is initializing.
+        window.backgroundColor = .windowBackgroundColor
         super.init(window: window)
 
         window.onPaste = { [weak self] in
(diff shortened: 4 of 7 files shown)


nesquena (approved):
Verified on Mac agent (branch at `c0206c3`):

- `swift build -c release` — Build complete (2.07s)
- `swift test` — 20/20 passed (PortValidationTests, SSHArgumentTests, URLValidationTests)

Info.plist keys (confirmed):
- `NSAppTransportSecurity` = `{ NSAllowsArbitraryLoads = false; NSAllowsArbitraryLoadsInWebContent = true }` — matches fix #25 (arbitrary loads off at top level; WKWebView can still load non-localhost `http://`)
- `SUPublicEDKey` = `daAlTqdBbYPSCDjS9IfTCJOFDo1jqtjMRZluhtAbKMY=`
- `SUFeedURL` = `https://hermes-webui.github.io/hermes-swift-mac/appcast.xml`

Initial build failed on a Swift access-level error at `BrowserWindowController.swift:30` — `var webViewForZoom: HermesWebView?` leaked a `private` type through an `internal` property. Fixed in `c0206c3` by narrowing the return type to `WKWebView?`, which is all the AppDelegate zoom handlers need (they only touch `.magnification`). Re-verified green after the fix.

Scope matches description: 7 files, 225 insertions on the feature commit. No scope creep. Ready to merge and tag v1.1.0.

PR #32, opened 2026-04-20
feat: v1.2.0 — notifications toggle, Cmd+R reload, Cmd+W hide, window memory, global hotkey Cmd+Shift+H

### LINKED ISSUE #6: feat: global keyboard shortcut to bring Hermes window forward from anywhere (Cmd+Shift+H)
## Overview

Register a system-wide keyboard shortcut (default: Cmd+Shift+H, user-configurable) that brings the Hermes window to the front from any app — without needing to click the Dock icon or use Cmd+Tab.

## Why

The app is a full-window Dock app (correct for a tool with sessions and scrollback), but getting back to it mid-task still requires a trip to the Dock or Cmd+Tab. A global shortcut removes that friction — invoke Hermes from VS Code, a browser, or a terminal without switching context.

## Behaviour spec

- Default shortcut: **Cmd+Shift+H**
- If the window is in the background: bring it to front and focus the chat input
- If already frontmost: optionally hide it (toggle)
- User-configurable in Preferences

## Implementation notes

- `NSEvent.addGlobalMonitorForEvents(matching:handler:)` — works for non-sandboxed apps without accessibility permission for modifier+key combos
- Action: `NSApp.activate(ignoringOtherApps: true)` + `browserWindow?.window?.makeKeyAndOrderFront(nil)` + focus chat input

## Note

Updated from original — this was previously scoped to the menu bar popover design (issue #2, now closed as wrong direction). A global shortcut makes just as much sense for the current full-window Dock app.

### LINKED ISSUE #28: ux: preferences notification copy says 'tab' — should say 'app' in native Mac context
## Problem

In Preferences → Browser Notifications, the description reads:

> "Show a system notification when a response completes while the **tab** is in the background."

This is a native Mac app — there is no browser, no tab. The word "tab" is browser-era copy that leaked in from the WebUI.

## Fix

Change to:

> "Show a system notification when a response completes while the **app** is in the background."

This is a one-word change in the WebUI's `i18n.js` (`settings_desc_notifications`) which is rendered inside the WKWebView. Should be fixed in both `nesquena/hermes-webui` (for browser users) and surfaced here for tracking.

*Cross-filed to nesquena/hermes-webui for the i18n string fix.*

nesquena-hermes:
## v1.2.0 — 6 QoL improvements

Sprint branch implementing the fixes from the [tonight's session brief](https://github.com/nesquena-hermes/hermes-webui-workspace/blob/main/docs/tonight-brief-swiftmac-apr20.md).

---

### What changed

**Fix 1 — Notification copy "tab" → "app" (fixes #28)**

Preferences now has a native notifications toggle with the correct copy: "Show a notification when a response completes while the **app** is in the background." Wired to `UserDefaults.standard.bool(forKey: "notificationsEnabled")` (default `true`). The notification handler in `BrowserWindowController` gates on this key. Window height bumped from 520 → 556 to fit the new row.

**Fix 2 — Cmd+, opens Preferences**

Already implemented in v1.0.0. Explicitly documented in CHANGELOG and noted here for completeness.

**Fix 3 — Cmd+R reloads page**

New View → Reload menu item with `keyEquivalent: "r"`. Handler calls `browserWindow?.webViewForZoom?.reload()` via the existing `webViewForZoom` accessor.

**Fix 4 — Cmd+W hides window instead of quitting**

`BrowserWindowController.windowShouldClose` returns `false` and calls `window?.orderOut(nil)`. `AppDelegate.applicationShouldTerminateAfterLastWindowClosed` changed to return `false`. `applicationShouldHandleReopen` added so clicking the Dock icon when the window is hidden re-surfaces it.

**Fix 5 — Window remembers last size + position**

`window.setFrameAutosaveName("HermesMainWindow")` in `BrowserWindowController.init`. NSWindow handles persistence automatically.

**Fix 6 — Global hotkey Cmd+Shift+H brings window forward (closes #6)**

Implemented using Carbon `RegisterEventHotKey` rather than `NSEvent.addGlobalMonitorForEvents`. 

> **Why Carbon, not NSEvent?** Opus confirmed that `NSEvent.addGlobalMonitorForEvents(matching: .keyDown)` requires Accessibility permission to fire when another app is frontmost — without it the handler silently does nothing from other apps. The brief's claim that no permission is required was incorrect. Carbon `RegisterEventHotKey` fires unconditionally at the event tap layer, with no user prompt and no app restart required.

The callback is a C closure that dispatches to main and calls `browserWindow?.showWindow(nil)` + `NSApp.activate(ignoringOtherApps: true)`. Self is retained via `Unmanaged.passRetained` and released in `applicationWillTerminate` alongside `UnregisterEventHotKey` and `RemoveEventHandler` cleanup.

---

### Files changed

- `Sources/HermesAgent/AppDelegate.swift` — notifications default seed, Cmd+R menu+handler, Cmd+W reopen handler, Carbon hotkey setup+teardown, `import Carbon.HIToolbox`
- `Sources/HermesAgent/BrowserWindowController.swift` — `windowShouldClose` hide override, `setFrameAutosaveName`, notification guard on `notificationsEnabled`
- `Sources/HermesAgent/PreferencesWindowController.swift` — notifications checkbox + handler, window height 520 → 556
- `CHANGELOG.md` — v1.2.0 entry

---

### Mac build verification (mandatory before merge)

Nathan — please run on the Mac agent:

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1.2.0 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

Expected: build complete, all tests pass, `bash build.sh` produces `Hermes Agent.app`.

Post an independent review from the `nesquena` account once green, and I'll merge + tag `v1.2.0`.

---

Closes #6, fixes #28.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [v1.2.0] — 2026-04-20
+
+### Fixed
+- **Notification copy** — preferences toggle now reads "Show a notification when a response completes while the **app** is in the background" — replacing the old "tab" wording left over from browser context. The toggle is wired to a new UserDefaults key (`notificationsEnabled`, default on) so users can disable native notifications. (fixes #28)
+
+### Added
+- **Cmd+,** — opens Preferences (standard macOS keyboard shortcut). Already wired; documented here explicitly.
+- **Cmd+R** — reloads the WebUI page via a new View → Reload menu item.
+- **Cmd+W** — hides the window instead of quitting. App stays running in the Dock; clicking the Dock icon brings it back. `applicationShouldTerminateAfterLastWindowClosed` returns `false` and `applicationShouldHandleReopen` re-surfaces the window.
+- **Window position memory** — `setFrameAutosaveName("HermesMainWindow")` restores last size and position on every launch.
+- **Global hotkey Cmd+Shift+H** — brings Hermes forward from any app using Carbon `RegisterEventHotKey`. No Accessibility permission required; works on first launch. (closes #6)
+
 ## [v1.1.0] — 2026-04-19
 
 ### Added
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -1,4 +1,5 @@
 import AVFoundation
+import Carbon.HIToolbox
 import Cocoa
 import Sparkle
 
@@ -19,6 +20,10 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     var preferencesWindow: PreferencesWindowController?
     var updaterController: SPUStandardUpdaterController!
 
+    // Global hotkey state (fix #6, Carbon-based — no Accessibility permission required)
+    private var carbonHotKeyRef: EventHotKeyRef?
+    private var carbonEventHandler: EventHandlerRef?
+
     func applicationDidFinishLaunching(_ notification: Notification) {
         NSApp.setActivationPolicy(.regular)
         NSApp.activate(ignoringOtherApps: true)
@@ -33,6 +38,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         setupMenu()
         seedDefaultsIfNeeded()
         requestMicrophonePermission()
+        setupGlobalHotkey()
         startTunnel()
     }
 
@@ -66,6 +72,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             defaults.set(defaultRemotePort, forKey: "remotePort")
             defaults.set(defaultTargetURL, forKey: "targetURL")
             defaults.set("direct", forKey: "connectionMode")
+            defaults.set(true, forKey: "notificationsEnabled")
         }
     }
 
@@ -253,6 +260,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let viewMenu = NSMenu(title: "View")
         viewMenuItem.submenu = viewMenu
         viewMenu.addItem(
+            withTitle: "Reload", action: #selector(reloadPage), keyEquivalent: "r")
+        viewMenu.addItem(.separator())
+        viewMenu.addItem(
             withTitle: "Zoom In", action: #selector(zoomIn), keyEquivalent: "+")
         viewMenu.addItem(
             withTitle: "Zoom Out", action: #selector(zoomOut), keyEquivalent: "-")
@@ -264,6 +274,12 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     @objc func checkForUpdates() {
         updaterController.checkForUpdates(nil)
+    }
+
+    // MARK: - Page reload (fix — Cmd+R)
+
+    @objc func reloadPage() {
+        browserWindow?.webViewForZoom?.reload()
     }
 
     // MARK: - View zoom (fix #24)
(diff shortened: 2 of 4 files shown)


nesquena (approved):
Verified on Mac agent at tip `998f73e` (feature `5da343d` + fix `998f73e`).

### Build/test results
- `swift build -c release` — Build complete (2.21s, 0 warnings)
- `swift test` — 20/20 passed

### Issues found and fixed in `998f73e`

1. **Build-blocker** in `AppDelegate.swift:324` — `InstallApplicationEventHandler` is a C function-like macro that Swift can't import. The CarbonEvents.h macro expands to `InstallEventHandler(GetApplicationEventTarget(), …)`; switched to that call form. Also swapped `Unmanaged.passRetained(self)` to `passUnretained` since `NSApp` owns its delegate for the full app lifetime and the handler is removed in `applicationWillTerminate` before teardown — `passRetained` leaked self at shutdown and the cleanup comment was wrong (no matching `Unmanaged.release()` existed).

2. **Upgrade-path regression** for `notificationsEnabled` — `seedDefaultsIfNeeded` gates on `sshUser == nil`, so it only ever runs on a brand-new install. Users upgrading from v1.1.0 (where notifications were always on) would read `bool(forKey: "notificationsEnabled") == false` and silently lose notifications. Fixed by registering a non-persistent default `["notificationsEnabled": true]` in `applicationDidFinishLaunching` — doesn't overwrite user-set values, gives upgraders the correct fallback.

3. **Preferences layout overlap** — window height bumped from 520→556 to make room for the new Notifications checkbox, but the starting `y = 460` was left unchanged. Result: Launch-at-Login checkbox pushed down to y=36 (span 36–58), overlapping the Save/Cancel buttons at y=16 (span 16–48). Fixed by bumping starting `y` from 460 to 496, which preserves the original spacing between Launch-at-Login and the button row.

4. **`var hkID` should be `let`** — minor compiler warning, fixed.

### Feature review

- **Notifications toggle** — guard in `BrowserWindowController.userContentController` correctly short-circuits before any notification-center calls. Auth cache on `notificationAuthGranted` avoids repeated authorization prompts. Good.
- **Cmd+R reload** — calls `webView.reload()` via the existing `webViewForZoom` accessor. Fine.
- **Cmd+W hides** — `windowShouldClose` returns false + `orderOut(nil)` + `applicationShouldTerminateAfterLastWindowClosed` returning false + `applicationShouldHandleReopen` bringing the window back are all consistent. Dock icon reopen works.
- **Window position memory** — `setFrameAutosaveName("HermesMainWindow")` is the one-line idiomatic approach.
- **Global hotkey Cmd+Shift+H (Carbon)** — Opus was right to flag `NSEvent.addGlobalMonitorForEvents` as needing Accessibility; Carbon `RegisterEventHotKey` fires without that prompt. Registration happens on main thread; handler dispatches back to main before touching AppKit. Signature `0x4845_524D = 'HERM'` is a valid unique four-char code. No `OSStatus` check on `RegisterEventHotKey` — if the combo is already registered by another app, the hotkey silently won't fire, but that's acceptable for now.

### Other observations (not blocking)
- `RegisterEventHotKey` return status isn't logged — if Cmd+Shift+H is already globally taken, users won't know why the shortcut is dead. Consider `NSLog` on non-zero status in a follow-up.
- Feature commit stats match the description: 4 files, 115 insertions. No scope creep on the branch itself.

Ready to merge and tag v1.2.0.

PR #33, opened 2026-04-20
fix: v1.2.1 — hotkey OSStatus check, redundant default seed, checkbox init

nesquena-hermes:
## v1.2.1 — three nits from the v1.2.0 review

All three items raised or implied by the v1.2.0 independent review. No behavior changes for users — the only observable difference is a diagnostic log line if Cmd+Shift+H conflicts.

---

### Changes

**1. RegisterEventHotKey OSStatus checked (reviewer's explicit note)**

```swift
// before
RegisterEventHotKey(UInt32(kVK_ANSI_H), UInt32(cmdKey | shiftKey), hkID,
    GetApplicationEventTarget(), 0, &carbonHotKeyRef)

// after
let status = RegisterEventHotKey(...)
if status != noErr {
    NSLog("[HermesAgent] RegisterEventHotKey failed (OSStatus %d) — Cmd+Shift+H may already be claimed by another app.", status)
}
```

**2. Redundant `notificationsEnabled` seed removed from `seedDefaultsIfNeeded`**

`UserDefaults.standard.register(defaults: ["notificationsEnabled": true])` in `applicationDidFinishLaunching` covers both fresh installs and upgrades. The duplicate `defaults.set(true, forKey: "notificationsEnabled")` inside `seedDefaultsIfNeeded` (which only runs on first launch) was redundant and slightly confusing. Removed.

**3. Notifications checkbox `target`/`action` set in initializer**

```swift
// before: init with nil then immediately reassign
notificationsCheckbox = NSButton(checkboxWithTitle: "...", target: nil, action: nil)
notificationsCheckbox.target = self
notificationsCheckbox.action = #selector(toggleNotifications(_:))

// after: set once, correctly
notificationsCheckbox = NSButton(checkboxWithTitle: "...", target: self,
    action: #selector(toggleNotifications(_:)))
```

---

### Mac build verification

Nathan — same drill as v1.2.0:

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1.2.1 && git pull --rebase
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1
```

Post approve from `nesquena` and I'll merge + tag `v1.2.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [v1.2.1] — 2026-04-20
+
+### Fixed
+- **RegisterEventHotKey OSStatus now checked** — if Cmd+Shift+H is already claimed by another app (Alfred, Raycast, etc.), a diagnostic `NSLog` fires instead of silently no-opping. Surfaced by the v1.2.0 independent review.
+- **Redundant `notificationsEnabled` seed removed** — `seedDefaultsIfNeeded()` was setting `notificationsEnabled` alongside `UserDefaults.standard.register(defaults:)`, which already covers both fresh installs and upgrades. The seed in `register(defaults:)` is the authoritative one; the duplicate in `seedDefaultsIfNeeded` is removed.
+- **Notifications checkbox init cleaned up** — `target` and `action` are now set in the `NSButton` initializer directly, removing the redundant two-line post-init reassignment.
+
 ## [v1.2.0] — 2026-04-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `9196fee`.

### Build/test results
- `swift build -c release` — Build complete (2.07s, 0 warnings)
- `swift test` — 20/20 passed

### Diff review — all three nits landed correctly

1. **`RegisterEventHotKey` OSStatus now checked** — captured into `let status`, `NSLog` fires on non-zero with the OSStatus value embedded. Message is self-explanatory when it shows in Console.app. This was exactly the follow-up I flagged on v1.2.0.

2. **Redundant `notificationsEnabled` seed removed from `seedDefaultsIfNeeded`** — correct. `register(defaults:)` in `applicationDidFinishLaunching` is the single source of truth now, handling both first-install and upgrade paths with one line. No behavioral regression: brand-new installs still see the checkbox as ON because `register(defaults:)` supplies the same `true` fallback that the persisted seed used to.

3. **Checkbox `target`/`action` set in initializer** — cleanly replaces `init(target: nil, action: nil)` + two reassignment lines with `init(target: self, action: #selector(...))`. Net -2 lines, identical runtime behavior.

Ready to merge and tag v1.2.1.

Issue #42, opened 2026-04-20
feat: multi-window or native tab bar support


nesquena-hermes:
## Feature request

Allow multiple independent Hermes sessions in separate tabs or windows — for example, one tab for a code project, another for a research thread, without losing context in either.

## Current behavior

The app is strictly single-window. `AppDelegate.browserWindow` holds one `BrowserWindowController`. Cmd+W hides it; there's no mechanism to open a second session.

## Why this is architecturally non-trivial

The app is a thin wrapper around a single hermes-webui URL. Each "session" is a browser session stored on the server side. Multiple tabs would require either:

a) Multiple WKWebView instances pointing at the same URL — sessions would be server-side and independent, but there's no multi-tab chrome (tab bar, tab switcher) in AppKit without significant custom UI work.

b) Multiple windows (simpler) — each `BrowserWindowController` is a full independent window. AppKit supports this naturally; the challenge is managing them all in AppDelegate (currently `browserWindow` is a single optional).

## Options

**Option A — Native tab bar (macOS Sierra+)**
`NSWindow` has built-in tab support via `tabbingMode`. Setting `.tabbingMode = .preferred` and implementing `NSWindowDelegate.window(_:willUseFullScreenPresentationOptions:)` gives you Safari-style native tabs essentially for free. Each tab would be a separate `NSWindow` in a tab group.

**Option B — Multiple independent windows**
Change `AppDelegate.browserWindow: BrowserWindowController?` to `AppDelegate.browserWindows: [BrowserWindowController]`. Add "New Window" (Cmd+N) to the menu. Simpler code, no tab chrome.

**Option C — In-app session switcher**
A sidebar or dropdown listing active sessions, implemented in JavaScript within the WKWebView. Would require hermes-webui cooperation. Out of scope for the native wrapper.

## Recommendation

Option A (native tab bar) is the lowest-effort path to multi-session and feels most Mac-native. Option B is useful for side-by-side comparison. Both are valid future directions.

## Scope note

This is a feature request for future consideration. The single-window architecture is intentional and correct for the current use case — this issue exists to capture the idea for evaluation, not to commit to building it.

Issue #41, opened 2026-04-20
feat: configurable global hotkey in Preferences


nesquena-hermes:
## Current behavior

The global hotkey is hardcoded to Cmd+Shift+H in `AppDelegate.setupGlobalHotkey()` (line ~322). The `RegisterEventHotKey` call uses `UInt32(kVK_ANSI_H)` and `UInt32(cmdKey | shiftKey)` with no user control. If Cmd+Shift+H is already claimed by Alfred, Raycast, or another app, the registration fails silently (now logged via NSLog since v1.2.1, but the user has no in-app recourse).

The shortcut is shown in Window → Show Hermes and in Preferences as a static label, but neither is interactive.

### UX pattern (standard macOS)

In Preferences, replace the static "Global shortcut: ⌘⇧H" label with a clickable recorder field:

```
Global shortcut:  [⌘⇧H]   ← click to change, then press new shortcut
```

- Clicking the field enters recording mode (highlighted bezel, text "Type shortcut…")
- User presses any key combination → field updates to show the new shortcut symbols
- Pressing Escape cancels; pressing Delete/Backspace clears the shortcut (disables global hotkey)
- On save, the new shortcut is stored in UserDefaults and `RegisterEventHotKey` is called with the new key code + modifiers

### Storage

```swift
// UserDefaults keys
"globalHotkeyKeyCode"    : Int   // e.g. kVK_ANSI_H = 4
"globalHotkeyModifiers"  : Int   // e.g. cmdKey | shiftKey
"globalHotkeyEnabled"    : Bool  // false = disabled (Delete clears it)
```

Default values (registered via `register(defaults:)` in `applicationDidFinishLaunching`):
```swift
"globalHotkeyKeyCode": kVK_ANSI_H,
"globalHotkeyModifiers": cmdKey | shiftKey,
"globalHotkeyEnabled": true
```

### Re-registration flow

When the user saves a new shortcut:
1. `UnregisterEventHotKey(carbonHotKeyRef)` — remove old registration
2. Store new values in UserDefaults
3. `RegisterEventHotKey(newKeyCode, newModifiers, ...)` — register new
4. Update the Window → Show Hermes menu item key equivalent to match

### Key recorder component

AppKit doesn't ship a built-in key recorder field. Options:
[...]

Issue #37, opened 2026-04-20
feat: find in page (Cmd+F) via WKWebView find API


nesquena-hermes:
## Current behavior

Pressing Cmd+F while the Hermes window is active does nothing. WKWebView does not wire browser-style find-in-page automatically — the key event falls through to AppKit with no handler, so users who instinctively reach for Cmd+F get silence.

## Why it matters

Hermes conversations get long. Searching for a specific response, code snippet, or mention is the natural reflex. The absence feels like a broken app rather than a missing feature.

## Proposed implementation

`WKWebView` exposes `find(_ string: String, configuration: WKFindConfiguration, completionHandler:)` (macOS 12+). The flow:

1. Add a **Find toolbar** or **Find panel** triggered by Cmd+F. Options:
   - **Minimal (recommended):** intercept Cmd+F in `BrowserWindow.performKeyEquivalent` (already subclassed in `BrowserWindowController.swift` lines 6–16 for paste). Show a native `NSSearchField` in a thin bar below the title bar. Wire it to `webView.find(...)` with `WKFindConfiguration` (case-insensitive, wraps). Cmd+G / Cmd+Shift+G advance/retreat results.
   - **Simpler but less native:** inject a JS `Ctrl+F` keydown event into the WebView and let the web app handle it — only works if hermes-webui has its own find UI, which it doesn't.
   - **Full AppKit integration:** implement `NSTextFinderClient` — a lot of boilerplate, probably overkill for a WKWebView wrapper.

2. `WKFindConfiguration` settings to consider:
   - `caseSensitive = false`
   - `wraps = true`
   - `backwards` toggled by Cmd+Shift+G

3. Dismiss bar on Escape; restore WebView first-responder status via existing `windowDidBecomeKey` pattern.

## Files to touch

- `Sources/HermesAgent/BrowserWindowController.swift` — find bar UI + `WKFindConfiguration` calls
- `Sources/HermesAgent/AppDelegate.swift` — View menu "Find…" item (Cmd+F)

## Scope

In: text search with next/previous navigation, result count label, Escape to dismiss.
Out: regex search, highlight-all toggle, search-in-selection. Those can come later.

## Availability

`WKWebView.find(_:configuration:completionHandler:)` requires macOS 12+. App already targets 12+, no `#available` guard needed.

PR #44, opened 2026-04-20
feat: v1.3.0 — auto-reconnect, zoom+fullscreen persistence, Dock badge, open-in-browser

### LINKED ISSUE #38: feat: auto-reconnect when network is restored (NWPathMonitor)
## Current behavior

When the network changes — laptop lid close/open, WiFi switch, LTE handoff, VPN connect — the app does nothing. In direct mode the health poll (30s timer in `BrowserWindowController.startHealthCheck()`) will eventually detect the backend is gone and update the window title dot, but there is no automatic recovery attempt. In SSH mode the tunnel process dies silently and the status dot turns red; the user must click Reconnect manually.

This is the most common real-world failure mode for SSH users specifically.

## Proposed implementation

Use `Network.framework`'s `NWPathMonitor` (macOS 10.14+, no import quirks) to watch for path changes.

**Where to add it:** `AppDelegate.swift`, initialized after `startTunnel()`.

```swift
import Network

private var pathMonitor: NWPathMonitor?
private var lastPathStatus: NWPath.Status = .satisfied

private func setupNetworkMonitor() {
    pathMonitor = NWPathMonitor()
    pathMonitor?.pathUpdateHandler = { [weak self] path in
        guard let self = self else { return }
        let wasUnsatisfied = self.lastPathStatus != .satisfied
        let nowSatisfied = path.status == .satisfied
        self.lastPathStatus = path.status
        // Only reconnect when going from no-network → network-restored.
        // Don't reconnect on every satisfied→satisfied path change (e.g. WiFi SSID switch
        // while still connected) to avoid spurious reconnect loops.
        if wasUnsatisfied && nowSatisfied {
            DispatchQueue.main.asyncAfter(deadline: .now() + 1.5) {
                self.startTunnel()
            }
        }
    }
    pathMonitor?.start(queue: DispatchQueue.global(qos: .background))
}
```

The 1.5s delay gives the new path time to stabilise before attempting a connection.

**Tear down** in `applicationWillTerminate`:
```swift
pathMonitor?.cancel()
pathMonitor = nil
```

## Phased approach

**Phase 1 (this issue):** auto-reconnect on network-restored. No UI changes.
**Phase 2 (future):** brief "Reconnected" toast or status update so the user knows it happened silently. Low priority.

## Edge cases to consider

- If the app is in error-window state (previous connect failed), `startTunnel()` already handles that path cleanly — it closes the error window and retries.
- If the user is mid-conversation and the network blips, `startTunnel()` will reload the WebView. This is acceptable — same as clicking Reconnect manually. A smarter future version could attempt a soft reload only if the backend responds.
- Rapid path thrashing (airplane mode toggle): the `wasUnsatisfied` guard prevents reconnect loops since we only fire when transitioning from unsatisfied → satisfied.

## Files to touch

- `Sources/HermesAgent/AppDelegate.swift` — add `import Network`, `pathMonitor` property, `setupNetworkMonitor()`, teardown in `applicationWillTerminate`

### LINKED ISSUE #39: feat: Dock icon badge when backend is unreachable
## Current behavior

The app's health state is only visible in the window title — a `●` (connected) or `○` (offline) dot next to the host name, implemented in `BrowserWindowController.updateWindowTitle(healthy:)` (line ~354). This is invisible when the window is hidden (Cmd+W) or minimised. A user who has put the app in the background has no OS-level signal that the backend has gone away.

## Proposed implementation

Use `NSApp.dockTile` to badge the Dock icon when the backend is unhealthy, and clear it when healthy.

```swift
// In BrowserWindowController.swift, inside updateWindowTitle(healthy:):
DispatchQueue.main.async {
    NSApp.dockTile.badgeLabel = healthy ? "" : "!"
}
```

That's literally it for the minimal version. `""` clears the badge; `"!"` shows a red exclamation dot on the Dock icon — the same pattern Mail uses for connection errors.

## Design options to consider

1. **`"!"` badge on error only (recommended for v1)** — minimal, clear, no false positives. Matches macOS convention for "attention needed."
2. **Green dot badge when healthy** — probably too noisy; users don't need constant confirmation the app is working.
3. **Custom `dockTile.contentImageSource`** — draw a tinted version of the app icon (red/green overlay). More polished but requires a `NSImageView` delegate. Future scope.

## Scope

This applies to **direct mode only** today — that's where `updateWindowTitle(healthy:)` is called (health poll runs every 30s). SSH mode uses `updateStatus()` which drives the in-window status bar; a future pass could also badge the Dock on SSH disconnect.

## Files to touch

- `Sources/HermesAgent/BrowserWindowController.swift` — two lines in `updateWindowTitle(healthy:)`

## Availability

`NSApp.dockTile` and `badgeLabel` have been available since macOS 10.5. No availability guards needed.

### LINKED ISSUE #43: feat: persist full-screen state and zoom level across launches
## Current behavior

Window size and position are restored correctly via `setFrameAutosaveName("HermesMainWindow")` (added in v1.2.0). However, **full-screen state is not preserved**. If the user:

1. Enters full-screen mode (green button or Cmd+Ctrl+F)
2. Quits the app
3. Relaunches

...the window opens in a regular windowed state instead of returning to full-screen. This is inconsistent with macOS convention — Safari, Terminal, and most AppKit apps restore full-screen state on relaunch.

## Why `setFrameAutosaveName` doesn't cover this

`NSWindow.setFrameAutosaveName` saves and restores the window frame (origin + size) but not the full-screen presentation state. Full-screen in AppKit is a separate mode managed by `NSWindowStyleMask.fullScreen` and `NSWindow.toggleFullScreen(_:)`, not encoded in the frame rect.

## Proposed implementation

**Save** full-screen state when it changes:

```swift
// In BrowserWindowController.swift — NSWindowDelegate callbacks
func windowDidEnterFullScreen(_ notification: Notification) {
    UserDefaults.standard.set(true, forKey: "windowWasFullScreen")
}

func windowDidExitFullScreen(_ notification: Notification) {
    UserDefaults.standard.set(false, forKey: "windowWasFullScreen")
}
```

**Restore** after the window has fully loaded and the WebView has connected (to avoid toggling full-screen during the splash/connect phase, which produces visual glitches):

```swift
// In BrowserWindowController — after first successful navigation
// i.e. in webView(_:didFinish:) or after health check confirms connected
private func restoreFullScreenIfNeeded() {
    guard UserDefaults.standard.bool(forKey: "windowWasFullScreen") else { return }
    window?.toggleFullScreen(nil)
}
```

The restoration should happen **after** the web content has loaded, not at window init — toggling full-screen on an empty WKWebView causes a jarring animation against a blank background.

## Additional persistence gap: zoom level

`webView.magnification` (set by Cmd+/Cmd−) is also not persisted. Each launch resets to 1.0. This is lower priority but worth noting — store in `UserDefaults.standard.double(forKey: "webViewZoom")` and restore in `buildUI()`.

## Files to touch

- `Sources/HermesAgent/BrowserWindowController.swift` — `windowDidEnterFullScreen`, `windowDidExitFullScreen`, `restoreFullScreenIfNeeded()` called from `webView(_:didFinish:)`

## Out of scope

Restoring the exact scroll position within the web app, or which conversation was open — that requires hermes-webui cooperation and is tracked separately.

nesquena-hermes:
## v1.3.0 — Connection resilience, state persistence, Dock feedback

Five features that make the app feel more polished and reliable, all additive and targeted.

**Find in page (Cmd+F)** was originally included but removed after Opus pre-commit review caught that `WKFindInteraction`/`isFindInteractionEnabled` are iOS-only symbols — they don't exist on macOS and would break the CI build. Filed separately as #44 with the correct NSTextFinder approach for macOS.

---

### What's in this PR

**Auto-reconnect on network restore** (closes #38)
`NWPathMonitor` watches for `.unsatisfied → .satisfied` path transitions. When the network comes back and the app is currently in an error or disconnected state, it automatically calls `startTunnel()` after a 1s debounce. Lid close/open, WiFi switch, VPN connect — all handled without user action. No permissions required. The loop guard is tight: if the backend is simply down but the network is healthy, the path stays `.satisfied` and no extra triggers fire.

**Zoom level persistence** (part of closes #43)
Zoom in/out/reset now saves `webView.magnification` to UserDefaults (`webViewMagnification` key) on every change. `webView(_:didFinishNavigation:)` restores it. The `0.5–3.0` range check filters out the unset `0.0` sentinel and invalid values.

**Full-screen state persistence** (part of closes #43)
`windowDidEnterFullScreen`/`windowDidExitFullScreen` delegate callbacks save a `windowWasFullScreen` Bool to UserDefaults. On the next launch, `openBrowser()` calls `toggleFullScreen(nil)` in a single-tick `DispatchQueue.main.async` block after `showWindow()`. The `isIntentionalClose` flag on `BrowserWindowController` prevents `windowDidExitFullScreen` from clobbering the saved preference when `startTunnel()` programmatically closes the window during a reconnect.

**Dock icon badge when offline** (closes #39)
`AppDelegate.setOfflineBadge(_:)` sets `NSApp.dockTile.badgeLabel` to `"!"` when the backend is unreachable, `nil` to clear. Triggered from four places: `showErrorWindow`, `openBrowser`, `BrowserWindowController.updateWindowTitle` (direct-mode health check), and `updateStatus` (SSH mode `.connected`/`.disconnected`). All calls dispatch to main internally. Visible even when the window is hidden with Cmd+W.

**View → Open in Browser** (bonus)
Opens the configured Hermes URL in the system default browser. 5 lines, zero risk. Useful for debugging or comparing sessions without changing Preferences.

---

### Issues addressed

- Closes #38 (auto-reconnect)
- Closes #39 (Dock badge)
- Implements #43 (zoom + full-screen persistence — both halves)

### Not in this PR

- #37 (Find in page) — blocked: `WKFindInteraction` is iOS-only. Needs `NSTextFinder` or JS injection. Filed as #44.
- #41 (configurable hotkey) — standalone Preferences work, separate sprint
- #42 (multi-window) — architectural, not a one-session feature

---

### Mac agent build verification (mandatory before merge)

Please run the following on the Mac agent and report the result:

```
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-3-0 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

Expected: `swift build` green, all tests pass, `build.sh` produces `Hermes Agent.app`.

No new Info.plist keys, no new entitlements, no new frameworks beyond `Network` (auto-linked, no Package.swift changes needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## [v1.3.0] — 2026-04-20
+
+### Added
+- **Auto-reconnect on network restore (NWPathMonitor)** — when WiFi drops and comes back,
+  or a VPN connects, the app automatically re-attempts connection without any manual click.
+  Only fires when the app is already in an error or disconnected state; no action is taken
+  if the backend is simply down with a healthy network. Uses `Network.framework`
+  `NWPathMonitor`, no Accessibility permission required. (closes #38)
+- **Zoom level persistence** — zoom level (Cmd++/Cmd+-/Cmd+0) is now saved to UserDefaults
+  and restored after every page load, including reconnects. (part of closes #43)
+- **Full-screen state persistence** — if the app was in full-screen when quit, it returns to
+  full-screen on next launch. Uses `NSWindowDelegate` `windowDidEnterFullScreen`/
+  `windowDidExitFullScreen` callbacks. (part of closes #43)
+- **Dock icon badge when offline** — the Dock icon shows a "!" badge when the backend is
+  unreachable (direct mode health check failure or SSH tunnel disconnects). Clears
+  automatically when the connection is restored. Visible even when the window is hidden. (closes #39)
+- **View → Open in Browser** — opens the configured Hermes URL in the system default browser.
+  Useful for comparison or debugging without changing any settings.
+
 ## [v1.2.2] — 2026-04-20
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -1,6 +1,7 @@
 import AVFoundation
 import Carbon.HIToolbox
 import Cocoa
+import Network
 import Sparkle
 
 class AppDelegate: NSObject, NSApplicationDelegate {
@@ -24,6 +25,12 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     private var carbonHotKeyRef: EventHotKeyRef?
     private var carbonEventHandler: EventHandlerRef?
 
+    // NWPathMonitor auto-reconnect (fix #38)
+    private let pathMonitor = NWPathMonitor()
+    private let pathMonitorQueue = DispatchQueue(label: "hermes.network.monitor")
+    private var lastPathStatus: NWPath.Status = .satisfied
+    private var pendingReconnect: DispatchWorkItem?
+
     func applicationDidFinishLaunching(_ notification: Notification) {
         NSApp.setActivationPolicy(.regular)
         NSApp.activate(ignoringOtherApps: true)
@@ -45,6 +52,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         requestMicrophonePermission()
         setupGlobalHotkey()
         startTunnel()
+        startPathMonitor()
     }
 
     private func requestMicrophonePermission() {
@@ -88,6 +96,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let splashSubtitle = connectionMode == "ssh" ? "Establishing SSH tunnel…" : "Connecting…"
         splashWindow = SplashWindowController(title: appTitle, subtitle: splashSubtitle)
         splashWindow.showWindow(nil)
+        browserWindow?.isIntentionalClose = true
         browserWindow?.close()
         browserWindow = nil
         errorWindow?.close()
@@ -172,9 +181,22 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         }
         browser.showWindow(nil)
         browserWindow = browser
+
+        // Restore full-screen state (fix #43)
+        if UserDefaults.standard.bool(forKey: "windowWasFullScreen") {
+            DispatchQueue.main.async {
+                if browser.window?.styleMask.contains(.fullScreen) == false {
+                    browser.window?.toggleFullScreen(nil)
+                }
+            }
+        }
+
+        // Clear offline badge when connected (fix #39)
+        setOfflineBadge(false)
     }
 
     private func showErrorWindow(targetURL: String, mode: String) {
+        browserWindow?.isIntentionalClose = true
         browserWindow?.close()
         browserWindow = nil
         let err = ErrorWindowController(
@@ -193,6 +215,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         err.showWindow(nil)
         err.window?.makeKeyAndOrderFront(nil)
         errorWindow = err
+
+        // Show offline badge when in error state (fix #39)
+        setOfflineBadge(true)
     }
 
     /// Verify the target URL answers HTTP before opening the main browser.
@@ -211,6 +236,44 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         URLSession.shared.dataTask(with: request) { _, response, _ in
             completion(response is HTTPURLResponse)
         }.resume()
+    }
+
+    // MARK: - Dock badge (fix #39)
+
+    func setOfflineBadge(_ offline: Bool) {
+        DispatchQueue.main.async {
+            NSApp.dockTile.badgeLabel = offline ? "!" : nil
+        }
+    }
+
+    // MARK: - NWPathMonitor auto-reconnect (fix #38)
+
+    private func startPathMonitor() {
+        pathMonitor.pathUpdateHandler = { [weak self] path in
+            guard let self = self else { return }
+            DispatchQueue.main.async {
+                let previous = self.lastPathStatus
+                self.lastPathStatus = path.status
+                // Only react to unsatisfied → satisfied transitions.
+                guard previous != .satisfied, path.status == .satisfied else { return }
+                self.scheduleAutoReconnect()
+            }
+        }
+        pathMonitor.start(queue: pathMonitorQueue)
+    }
+
+    private func scheduleAutoReconnect() {
+        pendingReconnect?.cancel()
+        let work = DispatchWorkItem { [weak self] in
+            guard let self = self else { return }
+            let inErrorState = self.errorWindow != nil
+                || self.tunnelManager?.status == .disconnected
+            guard inErrorState else { return }
+            NSLog("[HermesAgent] Network came back — auto-reconnecting")
+            self.startTunnel()
+        }
+        pendingReconnect = work
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1.0, execute: work)
     }
 
     func setupMenu() {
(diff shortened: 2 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `1920b90`.

### Build/test results
- `swift package resolve` — clean
- `swift build -c release` — Build complete (2.56s, 0 warnings)
- `swift test` — 20/20 passed

### Code review findings

**NWPathMonitor auto-reconnect (#38) — solid.**
- `lastPathStatus = .satisfied` initial value is correct: at launch, if network is up, first callback sees previous==current==satisfied and no-ops; if network is down, first callback flips to `.unsatisfied` without triggering (the guard requires `current == .satisfied`). Subsequent recovery transitions fire correctly.
- `pathUpdateHandler` fires on `pathMonitorQueue` and all mutable state (`lastPathStatus`, `pendingReconnect`) is touched only after dispatching to main — thread-safe.
- `pendingReconnect?.cancel()` + 1s debounce coalesces flapping WiFi events. The `inErrorState` gate (`errorWindow != nil || tunnelManager?.status == .disconnected`) is the tight-loop guard — a stray `.satisfied → .satisfied` would still no-op here even if something slipped past the first guard.
- Cleanup in `applicationWillTerminate` cancels both the monitor and any pending work item.

**Zoom persistence (#43) — correct.**
- Saved on every ⌘+/⌘-/⌘0 with the clamp already applied, so the stored value is always in [0.5, 3.0].
- Restored in `webView(_:didFinish:)` with a range check (`saved >= 0.5 && saved <= 3.0`) that correctly rejects the 0.0-on-unset sentinel.
- Minor nit (not blocking): the key `"webViewMagnification"` is a static on `AppDelegate` but a bare literal in `BrowserWindowController.webView(_:didFinish:)`. Keys in two places invites silent drift — worth promoting to a shared constant in a follow-up.

**Full-screen persistence (#43) — the `isIntentionalClose` flag is the key subtlety and it's handled right.**
- `windowDidExitFullScreen` fires when a window closes while in fullscreen. Without the flag, every reconnect path (`startTunnel`, `showErrorWindow`) would clobber `windowWasFullScreen = false` right before the app re-opens a new window. Both call sites correctly set `isIntentionalClose = true` before `close()`.
- User-initiated exit from fullscreen (green-button / ⌃⌘F) still flows through `windowDidExitFullScreen` with the flag false, so the preference updates — correct.
- Restore in `openBrowser` defers `toggleFullScreen` via `DispatchQueue.main.async` to let `showWindow` finish layout first. The `styleMask.contains(.fullScreen) == false` check is defensive but harmless (new window is never born in fullscreen).

**Dock badge (#39) — simple and well-placed.**
- `setOfflineBadge` bounces to main and sets `NSApp.dockTile.badgeLabel`.
- Invoked on `openBrowser` (clear), `showErrorWindow` (set), `updateStatus` for `.connected`/`.disconnected`, and `updateTitle` for healthy/unhealthy. Covers both SSH and direct paths plus background-window health changes.
- Minor coupling: `(NSApp.delegate as? AppDelegate)?.setOfflineBadge(...)` in `BrowserWindowController` assumes a concrete delegate type. Not a blocker since `NSApp.delegate` is set once at launch, but promoting the badge helper to a `ConnectionStatusObserver` protocol would be cleaner in a future refactor.

**Open in Browser — 5 lines, nothing to flag.**
Reads the configured `targetURL` (so a Preferences change is respected), falls back to `http://localhost:8787`, opens via `NSWorkspace.shared.open`.

**Dropped: Cmd+F find-in-page.** Correct call. `WKFindInteraction` / `isFindInteractionEnabled` are iOS/iPadOS 16+ only — they don't exist on macOS WKWebView. Would have been a compile-time CI break. The macOS-native path is `WKWebView.find(_:configuration:completionHandler:)` (macOS 12+) paired with a custom `NSSearchField` in a floating panel — worth a separate PR as #45 suggests.

### Non-blocking follow-ups
1. Extract `"webViewMagnification"` key into a single shared constant.
2. Consider a thin `ConnectionStatusObserver` protocol instead of casting `NSApp.delegate` in `BrowserWindowController`.
3. `NWPathMonitor` doesn't distinguish "internet up" from "backend reachable" — users still need the Reconnect button when the backend is the one that went away. Expected behavior; just worth documenting.

Ready to merge and tag v1.3.0.

PR #46, opened 2026-04-20
fix: v1.3.1 — shared zoomKey constant, NWPathMonitor comment, cast comment

nesquena-hermes:
## v1.3.1 — Reviewer follow-up nits from #44

Three non-blocking items flagged in the #44 review, addressed as a same-session patch:

**Shared zoom key constant**
`"webViewMagnification"` was `private static let zoomKey` in `AppDelegate` and a bare string literal in `BrowserWindowController.didFinishNavigation`. Made `AppDelegate.zoomKey` internal; `BrowserWindowController` now references `AppDelegate.zoomKey` directly. Single source of truth, no magic string duplication.

**NWPathMonitor reconnect scope comment**
Added inline comment to `scheduleAutoReconnect()` documenting that the method fires on network-link events (WiFi up, VPN connected), not on backend-health events. Future readers won't need to trace the call chain to understand why a healthy-network / down-backend scenario doesn't loop.

**`(NSApp.delegate as? AppDelegate)` cast comment**
Added a brief comment acknowledging the intentional coupling and why a `ConnectionStatusObserver` protocol abstraction isn't warranted for this single-delegate app. Documents the design decision explicitly.

No behaviour change. No new APIs. Swift build and all 20 tests should still be green.

### Mac agent build verification

```
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-3-1 && git pull --rebase
swift build -c release 2>&1
swift test 2>&1
```

Expected: same green result as #44.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [v1.3.1] — 2026-04-20
+
+### Fixed
+- **Shared zoom key constant** — `"webViewMagnification"` was defined as a private static
+  in `AppDelegate` and duplicated as a bare string literal in `BrowserWindowController`.
+  Made `AppDelegate.zoomKey` internal so `BrowserWindowController` references the single source
+  of truth. (reviewer follow-up from #44)
+- **NWPathMonitor reconnect scope documented** — added inline comment to `scheduleAutoReconnect`
+  clarifying that it fires on network-link events only, not backend-health events.
+- **`(NSApp.delegate as? AppDelegate)` cast comment** — acknowledged the intentional coupling
+  and why a full protocol abstraction isn't warranted here.
+
 ## [v1.3.0] — 2026-04-20
 
 ### Added
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -263,6 +263,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     private func scheduleAutoReconnect() {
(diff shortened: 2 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `686a52a`.

### Build/test results
- `swift build -c release` — Build complete (2.34s, 0 warnings)
- `swift test` — 20/20 passed

### Diff review — all three nits addressed cleanly

1. **`AppDelegate.zoomKey` promoted to `internal static`** (was `private static`). `BrowserWindowController.webView(_:didFinish:)` now reads `UserDefaults.standard.double(forKey: AppDelegate.zoomKey)` instead of the bare `"webViewMagnification"` literal. Single source of truth — silent drift is no longer possible. Access inside the AppDelegate itself (via `Self.zoomKey`) is unchanged.

2. **`scheduleAutoReconnect()` docstring** — clearly states "fires on network-link restoration (WiFi up, VPN connected, etc.), not on backend-health events. If the server is down but the network is healthy, no extra reconnect attempts fire — the path stays .satisfied so this is never called." That matches the actual behavior I traced in the #44 review. Good.

3. **`(NSApp.delegate as? AppDelegate)?` cast comment** — documents "The cast is safe — AppDelegate is always the app delegate in this single-delegate architecture. A ConnectionStatusObserver protocol would decouple this but adds boilerplate not warranted for a utility method." Reasonable trade-off call, now explicit in-code so future readers don't relitigate it.

### Behavior change
None. `internal static` vs `private static` is a visibility widening within the same module, not a semantic change. The two comments are pure documentation. 20/20 tests still green confirms no regression.

Ready to merge and tag v1.3.1.

PR #47, opened 2026-04-20
fix: v1.3.2 — mic permission no longer repeated, window size persists, nav failure guard

nesquena-hermes:
## v1.3.2 — Microphone permission fix, window size persistence, navigation guard

Three fixes, two user-reported.

---

### Fix 1 — Microphone permission prompt on every launch (user-reported)

**Root cause:** `requestMicrophonePermission()` was called unconditionally from `applicationDidFinishLaunching` on every launch. Once a user had denied mic access, this showed an `NSAlert` asking them to open System Settings on *every single launch*, even when the mic was never needed that session. The OS prompt also fired on every first-ever launch before the user had interacted with anything.

**Fix:** Deleted `requestMicrophonePermission()` entirely (21 lines) and removed its call from `applicationDidFinishLaunching`. Also removed the now-unused `import AVFoundation` from AppDelegate.

The `WKUIDelegate` method `requestMediaCapturePermissionFor` in `BrowserWindowController` already handles this correctly:
- On first mic use: asks the OS for permission, returns `.grant` or `.deny` based on the result
- On subsequent uses: if TCC is `.authorized`, returns `.grant` immediately — no prompt

The mic prompt will now only appear the first time the user actually clicks the mic button in the web UI. Once granted, it never prompts again.

### Fix 2 — Window size resets to 1280×830 on every launch (user-reported)

**Root cause:** `NSWindow.setFrameAutosaveName(_:)` saves frame changes to UserDefaults but does *not* restore a saved frame on window creation — that's `setFrameUsingName(_:)`. For windows created in Interface Builder, AppKit wires both automatically. For code-created windows (which this is), `setFrameUsingName` must be called explicitly.

**Fix:** Added `setFrameUsingName("HermesMainWindow")` immediately after `setFrameAutosaveName`. `center()` now only runs when `setFrameUsingName` returns `false` (first launch, no saved frame). Subsequent launches restore the exact last size and position.

```swift
// Before:
window.center()
window.setFrameAutosaveName("HermesMainWindow")

// After:
window.setFrameAutosaveName("HermesMainWindow")
if !window.setFrameUsingName("HermesMainWindow") {
    window.center()  // first launch only
}
```

### Fix 3 — Navigation failure double-fire guard

Both `didFailProvisionalNavigation` and the 5xx `decidePolicyFor navigationResponse` handler can fire on the same failing load event during teardown, opening the error window twice. Added `didReportNavigationFailure: Bool` flag; whichever handler fires first opens the error window, the second is a no-op.

---

### Issues / closes

- Fixes user-reported repeated microphone permission prompt
- Fixes user-reported window size reset on launch

---

### Mac agent build verification (required before merge)

```
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-3-2 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

No new Info.plist keys, no new entitlements, no new frameworks. `import AVFoundation` removed from AppDelegate (BrowserWindowController still imports it for `requestMediaCapturePermissionFor`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Changelog
 
+## [v1.3.2] — 2026-04-20
+
+### Fixed
+- **Microphone permission prompt no longer appears on every launch** — removed the proactive
+  `requestMicrophonePermission()` call from `applicationDidFinishLaunching`. That call fired
+  on every launch and, once the user had denied mic access, showed an `NSAlert` on *every*
+  subsequent launch regardless of whether the mic was needed. The `WKUIDelegate` method
+  `requestMediaCapturePermissionFor` already handles mic grants correctly and lazily — the
+  OS prompt only appears the first time the user actually clicks the mic button in the web UI.
+  (user-reported)
+- **Window size now persists across launches** — for programmatically created `NSWindow`
+  instances, `setFrameAutosaveName` saves future frame changes but does not restore the
+  previously saved frame on re-creation. Added `setFrameUsingName("HermesMainWindow")`
+  immediately after the autosave call; `center()` now only runs on first launch (when no
+  saved frame exists). Last used size and position are preserved across restarts. (user-reported)
+- **Navigation failure double-fire guard** — both `didFailProvisionalNavigation` and the
+  5xx `decidePolicyFor navigationResponse` handler can fire on the same failing load event
+  during teardown. Added `didReportNavigationFailure` flag so the error window is only
+  opened once per navigation attempt.
+
 ## [v1.3.1] — 2026-04-20
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -1,4 +1,3 @@
-import AVFoundation
 import Carbon.HIToolbox
 import Cocoa
 import Network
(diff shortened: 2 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `cdbe8b3`.

### Build/test results
- `swift test` — 20/20 passed

Scope: 35 additions / 24 deletions across 3 files. Matches the "three surgical changes" description.

### Code review findings

**Mic prompt removal — correct and net-improving.**
`requestMicrophonePermission()` was called unconditionally on every launch. In the `.denied` case it nagged with a full NSAlert every single time, even in sessions where the user never touched voice input. The replacement is zero code — WebKit's default `getUserMedia` → system-prompt path is already sufficient because all the required plumbing is in place:
- `NSMicrophoneUsageDescription` is written by both `build.sh:84` and `.github/workflows/build-release.yml:99` (parity preserved)
- `com.apple.security.device.microphone` entitlement is set in `Entitlements.plist:9`

With that, when `navigator.mediaDevices.getUserMedia({audio: true})` runs in the web app, WebKit surfaces the system mic prompt at the point of actual use. Denials then surface as a `NotAllowedError` to the frontend, which is a cleaner UX than the preemptive NSAlert. Removing `import AVFoundation` is the right follow-through — no remaining consumers.

**Window size persistence — correct diagnosis of an AppKit footgun.**
`setFrameAutosaveName` only wires up the *save* side; `setFrameUsingName` does the restore. AppKit wires both automatically for IB-loaded windows, but code-created windows need both explicitly. The new `if !setFrameUsingName("HermesMainWindow") { center() }` pattern:
1. Restores the saved frame on subsequent launches
2. Falls back to `center()` on first launch (when `setFrameUsingName` returns `false`)
3. Preserves the autosave wire-up so future frame changes continue to persist

No ordering concern — `setFrameAutosaveName` is called before `setFrameUsingName`, which is the correct order (the name must be bound before the restore can look it up).

**Navigation failure guard — defensive, single-shot, correctly scoped.**
The flag is on `BrowserWindowController`, so its lifetime is per-controller. When the user hits Retry and `AppDelegate.startTunnel()` creates a fresh controller, `didReportNavigationFailure` starts `false` again — no state leak between reconnect cycles.

Both guard sites are placed correctly:
- `didFailProvisionalNavigation` — the `NSURLErrorCancelled` early-return (for Safari-redirected links) runs *before* the guard, so those still don't consume the one-shot
- `decidePolicyFor navigationResponse` 5xx path — `decisionHandler(.cancel)` still fires unconditionally; only the `onNavigationFailed?()` invocation is gated

The race this protects against: a 5xx response arrives and we cancel the nav, then the underlying connection further drops and surfaces `didFailProvisionalNavigation` with a non-cancelled error code. Without the guard, `AppDelegate.showErrorWindow` runs twice — the second call leaks the first `ErrorWindowController` (because `errorWindow = err` overwrites the ref without closing the previous one) and the user briefly sees two error windows. Simple boolean, cheap fix.

### Non-blocking follow-ups
None this round. The three changes are each self-contained, each corrects a real bug, and together they remove more code than they add — a nice direction of travel.

Ready to merge and tag v1.3.2.

PR #48, opened 2026-04-20
fix: v1.3.3 — mic denied recovery alert, window frame via NSWindowController autosave

nesquena-hermes:
## v1.3.3 — Fix mic permission regression and window size regression from v1.3.2

Both bugs are regressions introduced in v1.3.2. Apologies for the churn.

---

### Fix 1 — Microphone "access denied" with no prompt (regression from v1.3.2)

**Root cause:** v1.3.2 correctly removed the aggressive launch-time mic prompt, but deleted the recovery path along with it. Users whose TCC permission was already `.denied` (from ever clicking "Don't Allow" — including on the v1.3.1 launch-time alert) hit the `.denied, .restricted` branch in `requestMediaCapturePermissionFor`, which called `decisionHandler(.deny)` and returned silently. No alert, no path to System Settings, mic permanently broken from the user's perspective.

**Fix:** The `.denied, .restricted` branch now:
1. Calls `decisionHandler(.deny)` first (so WebKit doesn't wait on a blocking modal)
2. Shows a single "Open System Settings" alert pointing to Privacy & Security → Microphone
3. Throttled via `private static var didShowMicDeniedAlert` — fires at most once per app session, and only for mic (not camera)

Fresh users (TCC `.notDetermined`) still hit the correct branch: `AVCaptureDevice.requestAccess` shows the OS prompt once, result is returned to WebKit, TCC remembers it permanently.

### Fix 2 — Window size still resetting to 1280×830 (regression from v1.3.2)

**Root cause:** `NSWindowController` has its own `windowFrameAutosaveName` property that governs frame persistence for the controller. v1.3.2 set `setFrameAutosaveName` and called `setFrameUsingName` on the raw `NSWindow` *before* `super.init(window:)`. When `NSWindowController` initialises, its default empty `windowFrameAutosaveName` overwrites the window-level setting — so subsequent resizes were registered under no name and never persisted to UserDefaults.

**Fix:** Removed the window-level calls entirely. After `super.init(window:)`, set `self.windowFrameAutosaveName = "HermesMainWindow"` on the controller. AppKit's controller machinery handles both save (on every resize/move) and restore (on window display) atomically, correctly, for the lifetime of the controller. `center()` is guarded to first-launch only by checking for the absence of the UserDefaults key `"NSWindow Frame HermesMainWindow"`.

---

### Mac agent build verification (required before merge)

```
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-3-3 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

No new Info.plist keys, no new entitlements, no new frameworks.

**Manual verification recommended:**
- Deny mic access in System Settings → Privacy & Security → Microphone, launch app, click mic → should see "Open System Settings" alert (once per session)
- Resize window, quit, relaunch → window should open at the same size and position

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## [v1.3.3] — 2026-04-20
+
+### Fixed
+- **Microphone "access denied" error after v1.3.2 upgrade** — v1.3.2 correctly removed the
+  aggressive launch-time mic prompt, but also inadvertently removed the recovery path for users
+  whose TCC permission was already `.denied`. Those users got silent failure with no way back.
+  Fixed: the `.denied` branch now calls `decisionHandler(.deny)` immediately (so WebKit doesn't
+  wait) and then shows a single "Open System Settings" alert pointing to Privacy & Security →
+  Microphone. Throttled to once per app session so it can't spam. (user-reported regression from v1.3.2)
+- **Window size still resetting to default after v1.3.2** — the v1.3.2 fix called
+  `setFrameAutosaveName` and `setFrameUsingName` on the raw `NSWindow` before `super.init`.
+  `NSWindowController` has its own `windowFrameAutosaveName` property that both saves and
+  restores the frame — when the controller initialises, its default empty value overwrote the
+  window-level setting, so subsequent resizes were never persisted. Fixed: removed the window-level
+  calls and set `self.windowFrameAutosaveName = "HermesMainWindow"` on the controller after
+  `super.init`. AppKit now manages save and restore atomically. `center()` is guarded to
+  first-launch only. (user-reported regression from v1.3.2)
+
 ## [v1.3.2] — 2026-04-20
 
 ### Fixed
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -631,7 +631,7 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
             DispatchQueue.main.async {
                 let alert = NSAlert()
                 alert.messageText = "Microphone Access Required"
-                alert.informativeText = "Hermes Agent needs microphone access for voice input. Enable it in System Settings \u2192 Privacy & Security \u2192 Microphone, then reload the page."
+                alert.informativeText = "Hermes Agent needs microphone access for voice input. Enable it in System Settings → Privacy & Security → Microphone, then reload the page."
                 alert.addButton(withTitle: "Open System Settings")
                 alert.addButton(withTitle: "Cancel")
                 if alert.runModal() == .alertFirstButtonReturn,
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -42,6 +42,9 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
     /// Guards against onNavigationFailed firing twice (both provisional and 5xx paths
     /// can trigger on the same load event during teardown).
     private var didReportNavigationFailure = false
+    /// Throttle the mic-denied alert to once per app session — avoids spamming if the
+    /// user hits the mic button multiple times after having denied access.
+    private static var didShowMicDeniedAlert = false
     /// Set to true before programmatic close so windowDidExitFullScreen
     /// doesn't clobber the saved full-screen preference (fix #43).
     var isIntentionalClose = false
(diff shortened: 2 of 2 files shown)


nesquena (approved):
Verified on Mac agent at tip `dd6bc37` (feature `76fc466` + fix `dd6bc37`).

### Build/test results
- `swift build -c release` — Build complete (2.20s, 0 warnings)
- `swift test` — 20/20 passed

Scope: 47/-9 across 2 files (CHANGELOG + BrowserWindowController) before the fix commit. Surgical.

### Issue found and fixed in `dd6bc37`

**Build-blocker: Swift Unicode escape syntax.** The v1.3.3 alert body used `\u2192` (C/JavaScript form, no braces):

```
alert.informativeText = "… System Settings \u2192 Privacy & Security \u2192 Microphone, …"
```

Swift requires `\u{XXXX}` with braces. Compiler diagnostic: `error: expected hexadecimal code in braces after unicode escape` at `BrowserWindowController.swift:634:152`. One-character fix: used the literal `→` arrow character, consistent with other user-facing strings in the codebase and with CHANGELOG entries that already use literal arrows.

### Code review findings (both root-cause fixes are correct)

**Mic denied recovery alert — correctly placed and correctly throttled.**
The v1.3.2 regression was that I approved a change that removed the launch-time `requestMicrophonePermission()` nag (good) but I missed that the `.denied` recovery path was only reachable *through* that launch-time nag — I was wrong in my #47 review to call it a complete deletion with no tradeoff. This PR restores the recovery path at the right location: inside the WKUIDelegate `.denied, .restricted` branch, triggered only when the user actually tries to use the mic. The ordering is correct:

1. `decisionHandler(.deny)` called synchronously first — WebKit can't wait on a modal alert, and blocking its completion handler would stall the WebKit process.
2. Guards: `!Self.didShowMicDeniedAlert` (throttle) AND `type != .camera` (don't fire the mic-specific alert for a pure camera request — `WKMediaCaptureType.cameraAndMicrophone` still triggers it, which is right because mic is part of the denial).
3. `Self.didShowMicDeniedAlert = true` — a type-level `static var`, so the flag survives across `BrowserWindowController` instances (reconnect cycles). That's the right scope for "once per app session" throttling.
4. `DispatchQueue.main.async` defers the `runModal()` past the WKUIDelegate callback frame — prevents reentrancy into WebKit while we have a modal up.
5. `"x-apple.systempreferences:com.apple.preference.security?Privacy_Microphone"` — correct deep-link URL for the exact settings pane the user needs.

If the user grants via System Settings and comes back: next `requestMediaCapturePermissionFor` sees `.authorized`, hits the `.grant` branch, alert never fires again. The stale `didShowMicDeniedAlert = true` doesn't matter because the denied branch is no longer hit.

**Window frame persistence via `NSWindowController.windowFrameAutosaveName` — correct diagnosis of the AppKit precedence rule.**
I reviewed and approved v1.3.2's `window.setFrameAutosaveName + setFrameUsingName` pair and didn't catch that NSWindowController has its own `windowFrameAutosaveName` property that *takes precedence* over anything set directly on the window. When the controller's own init runs inside `super.init(window:)`, it sees its empty `windowFrameAutosaveName` and wipes the binding set on the raw window. That's why frames never persisted in v1.3.2 despite the code looking right.

The new pattern is the correct one:
1. `super.init(window:)` first — the controller's init completes and its empty autosave name doesn't cause trouble because no user binding is in play yet.
2. `self.windowFrameAutosaveName = "HermesMainWindow"` on the controller — this is the authoritative setter, which tells AppKit both to save future changes and to restore the saved frame now.
3. `UserDefaults.standard.object(forKey: "NSWindow Frame HermesMainWindow") == nil` — correct first-launch detection (`"NSWindow Frame "` is AppKit's fixed prefix for the autosave UserDefaults key).
4. `window.center()` fallback — only on first launch; the subsequent `windowFrameAutosaveName` wire-up then captures that centered frame as the new saved default for future launches.

### Manual tests from the PR description
I can't run the interactive tests (deny mic → click mic in app → observe alert; resize → quit → relaunch → observe persistence) from the Mac agent build environment — those require running the `.app` with a live user at the keyboard. The alert logic and autosave wire-up were verified by code inspection against Apple's documented behavior (`WKMediaCaptureType`, WKUIDelegate threading, `NSWindowController.windowFrameAutosaveName`, the `"NSWindow Frame <name>"` UserDefaults key prefix).

### Non-blocking follow-ups

1. The autosave name `"HermesMainWindow"` appears twice in the frame-persistence code — once as `windowFrameAutosaveName` and once inside the UserDefaults key `"NSWindow Frame HermesMainWindow"`. A shared `static let windowFrameName = "HermesMainWindow"` would prevent silent drift if someone renames one but not the other. Same class of issue as the `zoomKey` duplication we fixed in v1.3.1.

Ready to merge and tag v1.3.3.

PR #49, opened 2026-04-20
fix: v1.3.4 — shared windowAutosaveName constant (follow-up from #48 review)

nesquena-hermes:
Reviewer follow-up from #48: `"HermesMainWindow"` appeared twice — as `windowFrameAutosaveName` and embedded in the derived UserDefaults key `"NSWindow Frame HermesMainWindow"`. Extracted to `private static let windowAutosaveName`. The derived key is now interpolated from the constant. Same pattern as `AppDelegate.zoomKey` fix in v1.3.1. Zero behavior change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,21 @@
 ## [v1.3.4] — 2026-04-20
 
 ### Fixed
+- **Microphone actually works again** — `getUserMedia()` was failing with `NotAllowedError`
+  even when TCC was `.authorized` (mic enabled in System Settings). Root cause: removing the
+  launch-time `requestMicrophonePermission()` call in v1.3.2 also removed its undocumented
+  side effect of initialising AVFoundation in the host process. The WebContent XPC process
+  has no `audio-input` entitlement of its own and inherits TCC attribution via the host's
+  active AVFoundation session. Without that session, capture fails at the platform layer
+  regardless of the delegate returning `.grant`. Fixed: `warmUpCaptureSubsystem()` calls
+  `AVCaptureDevice.default(for: .audio)` silently at launch — no UI, no prompt, no change
+  in UX — purely to establish the attribution chain that WebContent inherits.
+  (user-reported regression since v1.3.2)
 - **Shared window autosave name constant** — `"HermesMainWindow"` appeared twice in
   `BrowserWindowController`: once as the `windowFrameAutosaveName` value and once embedded
   in the derived UserDefaults key `"NSWindow Frame HermesMainWindow"`. Extracted to
   `private static let windowAutosaveName`. The derived key is now interpolated from the
-  constant, eliminating the drift risk. (reviewer follow-up from #48, same pattern as
-  `AppDelegate.zoomKey` fix in v1.3.1)
+  constant, eliminating the drift risk. (reviewer follow-up from #48)
 
 ## [v1.3.3] — 2026-04-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [v1.3.4] — 2026-04-20
+
+### Fixed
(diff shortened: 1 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `4dedc34`.

### Build/test results
- `swift test` — 20/20 passed

Scope: 38/-2 across 3 files. The branch contains two commits — `65417ce` (shared `windowAutosaveName` constant, addressing my v1.3.3 non-blocking follow-up) and `4dedc34` (the mic warm-up). Branch name and commit messages both read as v1.3.4.

### Code review

**`warmUpCaptureSubsystem()` — correct diagnosis, correct call, correct placement.**

The underlying behavior: WebKit's WebContent XPC renderer is a sandboxed Apple system process with no audio-input entitlement of its own. For `getUserMedia` inside the page to succeed, WebContent has to inherit TCC attribution from the host app — and on macOS 13+, "host app has TCC authorization" isn't sufficient; the host actually has to have initialized AVFoundation in-process so that WebContent's capture request can look up a live attribution chain. Without that warm-up, the capture request fails at the OS layer and surfaces as `NotAllowedError` in JS even when `AVCaptureDevice.authorizationStatus(for: .audio) == .authorized`.

This also explains why nobody noticed the bug before v1.3.2: the old `requestMicrophonePermission()` happened to call into AVFoundation every launch (via `AVCaptureDevice.authorizationStatus(for: .audio)` / `AVCaptureDevice.requestAccess`), which did the warm-up as a side effect. Deleting the method in v1.3.2 was the right UX call (no more launch-time nag) but accidentally deleted the load-bearing side effect. This PR restores just the side effect, without the nag.

The call `_ = AVCaptureDevice.default(for: .audio)`:
- Is the minimal warm-up that instantiates the AVFoundation audio singleton machinery in-process.
- Does *not* trigger any TCC prompt (that's `AVCaptureDevice.requestAccess`, a different entry point).
- Does *not* allocate a capture session or touch hardware.
- Returns `nil` gracefully if no audio device exists (headless CI/remote Macs), still initializes AVFoundation.
- The `_ =` discard is correct — the side effect is what matters.

Ordering is right: called in `applicationDidFinishLaunching` after `seedDefaultsIfNeeded()` and before `startTunnel()` — i.e. before any `BrowserWindowController` is created and therefore before any WebContent process is spawned. WebContent processes inherit attribution from the host at spawn time, so warming up before spawn is the correct temporal ordering.

The comment on the method is unusually good for this kind of non-obvious fix — future maintainers will be able to answer "why is this line here?" without git-archaeology.

**`windowAutosaveName` shared constant (65417ce) — clean refactor.**

`private static let windowAutosaveName = "HermesMainWindow"` on `BrowserWindowController`, referenced as `Self.windowAutosaveName` in both the `self.windowFrameAutosaveName = …` setter and the first-launch check `"NSWindow Frame \(Self.windowAutosaveName)"`. The `"NSWindow Frame "` prefix stays a literal — correct, since it's an AppKit-defined constant that won't change. Zero behavior change, satisfies the follow-up I flagged in the v1.3.3 review.

### Manual test
I can't run the interactive mic test (mic enabled in System Settings → click mic in hermes-webui → observe it works) from the Mac agent build environment. The fix was verified by code inspection against the documented AVFoundation/XPC attribution mechanism — the root-cause analysis in the PR body matches known macOS 13+ behavior.

### Non-blocking follow-ups
1. Consider a tiny regression test: assert that `AppDelegate.applicationDidFinishLaunching` doesn't return before `AVCaptureDevice.default(for: .audio)` is called. Easy to drop-in without needing microphone hardware. Worth it because this line is ~one word from being silently deleted again during a future refactor.

### PR metadata note
The PR title on GitHub still reads "shared windowAutosaveName constant (follow-up from #48 review)" — that was the original scope before the mic warm-up was force-pushed on top. Worth updating the title to reflect the mic fix since that's the load-bearing change; the shared constant became the smaller of the two.

Ready to merge and tag v1.3.4.

PR #51, opened 2026-04-20
fix: v1.3.6 — build.sh embeds entitlements in local builds, stale test comment

nesquena-hermes:
Reviewer follow-ups from #50: (1) `build.sh` codesign now passes `--entitlements Entitlements.plist` so local ad-hoc builds embed the same entitlements as CI DMGs — makes entitlement-gated features like mic testable locally without a CI run; (2) stale `default(for:)` reference in `LaunchBehaviorTests` updated to reflect the v1.3.5 root-cause fix. Zero behavior change to shipped DMGs (CI already had `--entitlements`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [v1.3.6] — 2026-04-20
+
+### Fixed
+- **`build.sh` now embeds entitlements in local ad-hoc builds** — `codesign` in `build.sh`
+  was signing without `--entitlements Entitlements.plist`, so locally-built `.app` bundles
+  never had any embedded entitlements (including `com.apple.security.device.audio-input`).
+  CI was correct (it already passed `--entitlements`). Local builds now embed the same
+  entitlements as CI-signed DMGs, making mic and other entitlement-gated features testable
+  without a full CI run. (reviewer follow-up from #50)
+- **Stale comment in `LaunchBehaviorTests`** — the warm-up regression history comment still
+  referenced `default(for:)` and omitted the v1.3.5 entitlement root cause. Updated to
+  accurately describe the fix history through v1.3.5. (reviewer follow-up from #50)
+
 ## [v1.3.5] — 2026-04-20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `98834b3`.

### Build/test results
- `swift build -c release` — Build complete (0.18s, 0 warnings)
- `swift test` — 22/22 passed

### Entitlement verification — now working on local builds

```bash
$ codesign -d --entitlements :- "Hermes Agent.app" 2>/dev/null
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0"><dict>
  <key>com.apple.security.device.audio-input</key><true/>
  <key>com.apple.security.network.client</key><true/>
</dict></plist>

$ codesign -d --entitlements :- "Hermes Agent.app" 2>/dev/null | grep audio-input
... com.apple.security.device.audio-input ...
```

Both entitlements from `Entitlements.plist` are now embedded in the locally-built `.app`. This matches what CI has been producing all along — local builds and production DMGs now have identical signing behavior for entitlements. The mic fix from v1.3.5 is now reproducible in local dev testing.

### Diff review — exactly the two follow-ups I flagged on #50

1. **`build.sh:107`** — one-line addition of `--entitlements Entitlements.plist` to the main app codesign call. Framework signing (line 105, Sparkle) stays entitlements-free, which is correct: frameworks don't carry their own entitlements. The comment block above the signing section explains the behavior and correctly notes that ad-hoc signing doesn't verify entitlements (only notarized/production signing does), so this is about parity with CI rather than a security guarantee.

2. **`LaunchBehaviorTests.swift:23–26`** — comment refresh. The old text claimed mic was "restored as a silent warm-up in v1.3.4 (PR #49)" which was incomplete (v1.3.4's warm-up was `default(for:)` — IOKit-only — which didn't actually fix the problem). The new text correctly walks the v1.3.2–v1.3.5 trail, identifies the entitlement rename as the real root cause, and documents the `requestAccess` vs `default(for:)` distinction. This matches the actual codebase as of v1.3.5.

### Zero behavior change to shipped DMGs
CI already had `--entitlements Entitlements.plist` at `.github/workflows/build-release.yml:182`, so the signed-and-notarized artifact is byte-identical before and after this PR for entitlement content. The only change is that local developer builds now look the same as production.

### Non-blocking follow-ups
None this round. This is a pure housekeeping PR that closes the two loops I opened in the v1.3.5 review.

Ready to merge and tag v1.3.6.

Issue #40, opened 2026-04-20
feat: improve first-launch and connection-failure experience


nesquena-hermes:
### Splash screen (`SplashWindowController.swift`)
Shows the app title, a spinner, and either "Connecting…" or "Establishing SSH tunnel…". No context about what the app is or what it needs to work. A brand-new user who downloaded the DMG without reading the README sees this for 4 seconds (the `preflightHTTP` timeout) and then hits the error screen.

### Error screen (`ErrorWindowController.swift`)
Direct mode shows:
- "Cannot connect to Hermes"
- The target URL (`http://localhost:8787`)
- "Make sure hermes-webui is running on the target URL. Start it with: `cd ~/hermes-webui-public && bash start.sh`"
- A link to `github.com/nesquena/hermes-webui`

Problems:
1. `~/hermes-webui-public` is a hardcoded path that won't match most users' actual install location.
2. The message assumes the user already knows what hermes-webui is. A new user doesn't.
3. There's no mention of Docker (a common install path), no link to the Quick Start, and no way to know if the issue is "server not running" vs "server running on a different port."
4. The "Don't have it yet?" label + repo link are on the same line and easy to miss — they're positioned at y=108–110 with no visual separation from the error body.

### Phase 1 — Error screen copy and layout (low effort, high impact)

**Distinguish first-time vs returning user:**

Check `UserDefaults.standard.bool(forKey: "hasConnectedBefore")` — set to `true` the first time a WebView load succeeds (in `webView(_:didFinish:)` in `BrowserWindowController`).

- **First-time user** (never connected): show more welcoming copy focused on setup.
- **Returning user** (has connected before): keep the current terse "server not running" message — they know the drill.

**First-time error copy:**
```
Hermes couldn't find your AI server.

Hermes is a frontend — it needs a running hermes-webui server to connect to.
The default address is http://localhost:8787.

Quick options:
  • Run locally   →  cd /path/to/hermes-webui && bash start.sh
  • Use Docker    →  docker run -p 8787:8787 nesquena/hermes-webui
  • Remote server →  open Preferences and configure an SSH tunnel

[Get hermes-webui]  [Open Preferences…]  [Try Again]
```

Replace the hardcoded `~/hermes-webui-public` path with a generic instruction. Add a Docker option. Make the repo link a proper button, not an inline afterthought.

**Returning user error copy:**
```
Cannot connect to Hermes
http://localhost:8787

Your server doesn't appear to be running.
[...]

PR #53, opened 2026-04-23
fix: eliminate startup white flash — show window after first WKWebView paint

### LINKED ISSUE #52: bug: white flash on startup still present — window shown before WKWebView first paint
## Bug description

On launch, there's a brief white flash between the splash window closing and the browser window fully rendering. The window appears before WKWebView has painted its first frame.

**Still present in v1.3.6** despite the v1.1.0 fix attempt (which set `underPageBackgroundColor` and injected a dark-mode `documentStart` script).

## Root cause

The v1.1.0 fix addressed the WKWebView's own backing layer but missed the real source of the flash. The launch sequence is:

```
splashWindow.close()
browser.showWindow(nil)   ← window is now visible, WKWebView hasn't painted
WKWebView starts loading  ← HTTP round trip, 100–500ms
Page renders
```

During that gap the native NSWindow frame is visible with its unrendered background. In light mode, `.windowBackgroundColor` is white — so light mode users always see it. The `documentStart` dark-mode script only fires for dark mode users and still doesn't cover this window-frame gap.

## Fix

Start the browser window invisible (`alphaValue = 0`), then fade it in on the first `didFinishNavigation` callback — the point where WKWebView has actually painted the page. This eliminates the flash entirely regardless of color scheme.

Two-line change across two files:
- `AppDelegate.swift` — set `browser.window?.alphaValue = 0` immediately after `browser.showWindow(nil)`
- `BrowserWindowController.swift` — in `didFinishNavigation`, fade from 0→1 on first load only

If `didFailProvisionalNavigation` fires instead (server unreachable), the window is closed via `onNavigationFailed` anyway — no orphaned invisible window.

nesquena-hermes:
## Summary

Eliminates the white flash on startup that's been present since v1.0.0 and survived the v1.1.0 fix attempt.

## Root cause

The v1.1.0 fix set `underPageBackgroundColor = .windowBackgroundColor` and injected a dark-mode `documentStart` script. Neither addressed the actual source of the flash. The window becomes visible via `showWindow(nil)` before WKWebView has made its first paint — the native NSWindow frame shows with an unrendered background. In light mode `.windowBackgroundColor` is white, so light mode users always see it. The dark-mode script didn't help them at all.

## Fix

Start the browser window invisible (`alphaValue = 0`), then fade it to opaque in `didFinishNavigation` on the first successful paint. Two changes:

**`AppDelegate.swift`**
```swift
browser.showWindow(nil)
browser.window?.alphaValue = 0  // hidden until first paint (fix #52)
```

**`BrowserWindowController.swift` — `didFinishNavigation`**
```swift
if window?.alphaValue == 0 {
    NSAnimationContext.runAnimationGroup { ctx in
        ctx.duration = 0.15
        window?.animator().alphaValue = 1
    }
}
```

**`BrowserWindowController.swift` — `didFailProvisionalNavigation`**
```swift
window?.alphaValue = 1  // restore before error window takes over
```

The `alphaValue == 0` guard means subsequent navigations (SPA route changes, Cmd+R reloads) are unaffected — they see alpha=1 and skip the animation entirely.

If navigation fails before the first paint, `didFailProvisionalNavigation` restores alpha before `onNavigationFailed` closes the window — no orphaned invisible window.

Closes #52.

---

**Note for reviewer:** This requires a Swift build to verify — please ask Nathan to run `swift test` and do a quick smoke test on the Mac.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **White flash on startup eliminated** — the window was made visible via `showWindow(nil)` before WKWebView had painted its first frame, causing a white flash on any color scheme (not just dark mode). The v1.1.0 fix set `underPageBackgroundColor` but missed this window-level gap. Fix: `alphaValue` is set to `0` before `showWindow`, then faded to `1` in `didFinishNavigation` on the first successful paint (0.15s animation). `didFailProvisionalNavigation` also restores alpha before handing off to the error window. Closes #52. (`AppDelegate.swift`, `BrowserWindowController.swift`)
+
 ## [v1.3.6] — 2026-04-20
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -160,6 +160,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             browser.updateStatus(tunnelManager.status, host: host, port: port)
         }
         browser.showWindow(nil)
+        browser.window?.alphaValue = 0  // hidden until first paint (fix #52)
         browserWindow = browser
 
         // Restore full-screen state (fix #43)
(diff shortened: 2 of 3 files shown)


nesquena-hermes:
Superseded by #54 which includes a more thorough fix: sets `alphaValue=0` before `showWindow` (correct ordering), uses `hasCompletedFirstPaint` flag instead of `alphaValue==0` check, adds dark background at all three layers (NSWindow, underPageBackgroundColor, documentStart script), and also bundles Cmd+F find bar (#37/#45) and error UX cleanup (#40) in the same v1.4.0 sprint.

PR #54, opened 2026-04-23
feat: v1.4.0 — white flash fix, Cmd+F find bar, error UX cleanup

### LINKED ISSUE #37: feat: find in page (Cmd+F) via WKWebView find API

### LINKED ISSUE #40: feat: improve first-launch and connection-failure experience

### LINKED ISSUE #45: feat: find in page (Cmd+F) — needs NSTextFinder or JS injection (not WKFindInteraction — iOS only)
## What happened

`WKFindInteraction` and `isFindInteractionEnabled` are iOS-only WebKit symbols — they don't exist on macOS. This was caught by Opus pre-commit review during the v1.3.0 sprint and removed before the PR was pushed (it would have broken the CI build with a compile error).

## What the correct macOS approach is

macOS WKWebView doesn't expose a native find bar API directly. The options for a non-sandboxed AppKit app are:

**Option A — JavaScript injection** (simplest)
Inject a lightweight JS find bar (highlight all, prev/next) via `WKUserScript`. No AppKit API needed. Works on macOS 12+. Downsides: custom UI, doesn't match system look.

**Option B — NSTextFinder + WKWebView** (system look, complex)
Implement `NSTextFinderClient` protocol on a custom NSView subclass wrapping WKWebView, then connect it to `NSTextFinder`. The text finder provides the standard macOS find bar. Significantly more code — WKWebView doesn't expose its text content to NSTextFinder out of the box, so you have to bridge via JS evaluation.

**Option C — evaluateJavaScript with `window.find()`**
Use `webView.evaluateJavaScript("window.find(\"\(query)\")")` after presenting a native `NSSearchField`-based sheet. The browser's built-in text search without a native find bar.

## Recommendation

Option A (JS injection) is the fastest path to a v1.3.1 release. Option B would be the most native-feeling but is a multi-session effort.

## References

- Opus advisory: `WKFindInteraction` header guarded by `#if TARGET_OS_IOS || TARGET_OS_VISION` — completely absent from macOS SDK
- Removed from PR #44 before push

### LINKED ISSUE #52: bug: white flash on startup still present — window shown before WKWebView first paint

nesquena-hermes:
## Summary

v1.4.0 sprint: white flash fix (for real), Cmd+F find in page, and error UX cleanup.

---

### Fix: white flash on startup — finally eliminated (closes #52)

Three-layer defence:

1. **`alphaValue = 0` before `showWindow`** in AppDelegate — the window is invisible when it first appears. No dark or white flash of any kind.
2. **Fade `0 → 1` (0.15s) in `didFinishNavigation`** — fires when WKWebView has actually painted its first frame. A `hasCompletedFirstPaint` bool flag ensures the animation fires exactly once; SPA route changes and Cmd+R reloads are unaffected.
3. **Dark background everywhere** — `NSWindow.backgroundColor`, `WKWebView.underPageBackgroundColor`, and the `documentStart` script all use `#1a1a1a` (not `windowBackgroundColor`, which is white in light mode). Any transient gap before the fade-in is dark regardless of system colour scheme.

If navigation fails before the first paint, `didFailProvisionalNavigation` restores `alphaValue = 1` before the error window takes over — no orphaned invisible window.

Previous fix attempts (v1.1.0: `underPageBackgroundColor` + dark-mode-only `documentStart` script) missed the NSWindow frame gap entirely and didn't help light mode users at all.

### Add: find in page — Cmd+F (closes #37, closes #45)

Pressing Cmd+F opens a native vibrancy bar (NSVisualEffectView, `.headerView` material) below the title bar with an NSSearchField. Results cycle via `window.find()` JS:

- Cmd+F: toggle find bar
- Cmd+G / `›` button: find next
- Cmd+Shift+G / `‹` button: find previous
- Done button or Escape: close

The Cmd+V paste intercept in `BrowserWindow.performKeyEquivalent` is now gated on `!(firstResponder is NSText)` so pasting into the find field works correctly.

Note: `WKFindInteraction` is iOS-only and won't compile on macOS — `window.find()` via JS evaluation is the correct macOS approach (as documented in #45).

### Change: error screen copy no longer hardcodes `~/hermes-webui-public` (closes #40)

Changed "Start it with: `cd ~/hermes-webui-public && bash start.sh`" to the more general "Run: bash start.sh (or: docker compose up -d)" which doesn't assume any specific install path and covers Docker users.

---

## Files changed

| File | What changed |
|------|-------------|
| `AppDelegate.swift` | Set `alphaValue = 0` before `showWindow` |
| `BrowserWindowController.swift` | Dark bg, fade-in on first paint, Cmd+F find bar, Cmd+V gate fix |
| `ErrorWindowController.swift` | Remove hardcoded `~/hermes-webui-public` path |
| `CHANGELOG.md` | v1.4.0 entry |

---

## Pre-merge checklist (for Mac agent)

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1-4-0 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

Expected: build complete, all tests pass, bash build.sh produces Hermes Agent.app.

**Smoke test:** launch the built app, verify:
1. No white or dark flash on startup (window fades in cleanly)
2. Cmd+F opens the find bar, Escape closes it, Cmd+G cycles results
3. Cmd+V in the find field pastes text (not routed to webView)

Closes #52, closes #37, closes #45, closes #40.
Supersedes #53 (which will be closed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## [v1.4.0] — 2026-04-23
+
+### Fixed
+- **White flash on startup eliminated (for real this time)** — the window is now started
+  with `alphaValue = 0` and fades in (0.15s) on `didFinishNavigation` — the point at which
+  WKWebView has actually painted its first frame. A `hasCompletedFirstPaint` flag ensures
+  the animation only fires once; SPA route changes and Cmd+R reloads are unaffected.
+  Additionally, the NSWindow frame background and the WKWebView pre-paint `documentStart`
+  script are both set to `#1a1a1a` (dark) regardless of the system colour scheme, so any
+  gap between window-visible and first paint is never white in any situation. Closes #52.
+
+### Added
+- **Find in page — Cmd+F** — pressing Cmd+F opens a native find bar (NSSearchField overlay
+  with vibrancy) anchored below the title bar. `‹` / `›` buttons and Cmd+G / Cmd+Shift+G
+  cycle through results via `window.find()` JS. Pressing Done or Escape closes the bar.
+  Closes #37, closes #45.
+
+### Changed
+- **Connection error copy no longer hardcodes `~/hermes-webui-public`** — the error
+  window now says "Run: bash start.sh (or: docker compose up -d)" instead of the path
+  that only applied to one specific install layout. Closes #40.
+
 ## [v1.3.6] — 2026-04-20
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -159,6 +159,10 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         if mode == "ssh", let host = sshHost, let port = localPort {
             browser.updateStatus(tunnelManager.status, host: host, port: port)
         }
+        // Fix #52: set alphaValue=0 BEFORE showWindow — prevents a brief
+        // visible-at-full-opacity tick. The window fades in on first WKWebView paint.
+        // backgroundColor is already set to #1a1a1a in BrowserWindowController.init.
+        browser.window?.alphaValue = 0
         browser.showWindow(nil)
         browserWindow = browser
 
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -5,10 +5,34 @@ import WebKit
 
 class BrowserWindow: NSWindow {
     var onPaste: (() -> Void)?
+    var onFind: (() -> Void)?
+    var onFindNext: (() -> Void)?
+    var onFindPrev: (() -> Void)?
 
     override func performKeyEquivalent(with event: NSEvent) -> Bool {
-        if event.modifierFlags.contains(.command) && event.charactersIgnoringModifiers == "v" {
+        // Cmd+V: route to the web view paste handler — but NOT when a native
+        // text field (e.g. the find bar's NSSearchField) is focused.
+        if event.modifierFlags.intersection(.deviceIndependentFlagsMask) == .command,
+           event.charactersIgnoringModifiers == "v",
+           !(firstResponder is NSText) {
             onPaste?()
+            return true
+        }
+        // Cmd+F: open find bar (fix #37/#45)
+        if event.modifierFlags.intersection(.deviceIndependentFlagsMask) == .command,
+           event.charactersIgnoringModifiers == "f" {
+            onFind?()
+            return true
+        }
+        // Cmd+G: find next; Cmd+Shift+G: find previous (fix #37/#45)
+        if event.modifierFlags.intersection(.deviceIndependentFlagsMask) == .command,
+           event.charactersIgnoringModifiers == "g" {
+            onFindNext?()
+            return true
+        }
+        if event.modifierFlags.intersection(.deviceIndependentFlagsMask) == [.command, .shift],
+           event.charactersIgnoringModifiers == "G" {
+            onFindPrev?()
             return true
         }
         return super.performKeyEquivalent(with: event)
@@ -42,6 +66,13 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
     /// Guards against onNavigationFailed firing twice (both provisional and 5xx paths
     /// can trigger on the same load event during teardown).
     private var didReportNavigationFailure = false
+    /// Tracks whether the first navigation paint has occurred, so the fade-in
+    /// animation (fix #52) only fires once — not on every SPA route change.
+    private var hasCompletedFirstPaint = false
+    // Find bar (fix #37/#45)
+    private var findBar: NSView?
+    private var findField: NSSearchField?
+    private var findBarVisible = false
     /// The UserDefaults autosave name for the main window frame.
     /// Used for both windowFrameAutosaveName and the derived "NSWindow Frame <name>" key.
     private static let windowAutosaveName = "HermesMainWindow"
@@ -69,9 +100,11 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
             defer: false
         )
         window.title = title
-        // Fix #23: set native window background to dark before content loads,
-        // so there's no white frame visible while WKWebView is initializing.
-        window.backgroundColor = .windowBackgroundColor
+        // Fix #23 + #52: set native window background to a dark colour before
+        // content loads. Using a literal dark value (not windowBackgroundColor)
+        // ensures the gap between window-visible and first-paint is dark in
+        // *all* colour schemes — windowBackgroundColor is white in light mode.
+        window.backgroundColor = NSColor(red: 0.10, green: 0.10, blue: 0.10, alpha: 1.0)
         super.init(window: window)
 
         // Persist and restore window frame across launches.
@@ -87,6 +120,15 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
 
         window.onPaste = { [weak self] in
             self?.handlePaste()
+        }
+        window.onFind = { [weak self] in
+            self?.toggleFindBar()
+        }
+        window.onFindNext = { [weak self] in
+            self?.findNext(forward: true)
+        }
+        window.onFindPrev = { [weak self] in
+            self?.findNext(forward: false)
         }
         window.delegate = self
 
@@ -186,18 +228,21 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         // before the first paint — otherwise the WebView renders white while
         // the dark theme is still loading from the server.
         if #available(macOS 12.0, *) {
-            webView.underPageBackgroundColor = .windowBackgroundColor
+            // Fix #52: match the pre-paint dark background so the overscroll gutter
+            // is dark too — not white, regardless of system colour scheme.
+            webView.underPageBackgroundColor = NSColor(red: 0.10, green: 0.10, blue: 0.10, alpha: 1.0)
         }
-        // Fallback for older OS: inject a documentStart script that sets the
-        // background immediately before the first paint.
+        // Fix #52: inject a documentStart script that always sets the document
+        // background to match our pre-paint NSWindow dark background (#1a1a1a),
+        // regardless of colour scheme. This eliminates any FOUC during the HTTP
+        // round-trip — even if the window becomes visible early, both the native
+        // frame and the WebView show the same dark colour.
+        // Once the app's actual CSS loads, it overrides this with the correct theme.
         let darkModeScript = WKUserScript(
             source: """
                 (function() {
-                    var isDark = window.matchMedia('(prefers-color-scheme: dark)').matches;
-                    if (isDark) {
-                        document.documentElement.style.background = '#1a1a1a';
-                        document.body && (document.body.style.background = '#1a1a1a');
-                    }
+                    document.documentElement.style.background = '#1a1a1a';
+                    if (document.body) { document.body.style.background = '#1a1a1a'; }
                 })();
                 """,
             injectionTime: .atDocumentStart,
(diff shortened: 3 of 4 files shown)


nesquena (approved):
Verified on Mac agent at tip `58e100e`.

### Build/test results
- `swift package resolve` — Sparkle 2.9.1 resolved
- `swift build -c release` — Build complete (5.03s after clean, 0 warnings)
- `swift test` — 22/22 passed (including the two `LaunchBehaviorTests` documentation stubs from v1.3.5)
- `codesign -d --entitlements :-` — `audio-input` + `network.client` both embedded (v1.3.6 carry-forward holds)

Scope: 220/-14 across 4 files. Matches the PR description: three features (white flash fix, find bar, error UX cleanup) in one commit.

### Code review

**Fix #52 — white flash on startup. Three rendering layers unified on `#1a1a1a`, plus fade-in.**

Traced the flash-path fix end to end — the dark value has to agree across every layer that can be visible before the app's CSS loads, otherwise one of them flashes white:

1. **`NSWindow.backgroundColor` = `NSColor(red: 0.10, green: 0.10, blue: 0.10)`** — this is `#1a1a1a` (0.10 × 255 ≈ 25.5 → 0x1A). Fixes the native frame. The old `windowBackgroundColor` was system-adaptive and went white in light mode — that's why the original fix was incomplete.
2. **`WKWebView.underPageBackgroundColor` = same literal `#1a1a1a`** — fixes the overscroll gutter.
3. **`document.documentElement.style.background = '#1a1a1a'`** unconditional (was: gated on `prefers-color-scheme: dark`). Fixes the body background during the HTTP round-trip, regardless of system colour scheme.
4. **`alphaValue = 0` before `showWindow`**, then animated to 1 on first `didFinish` — the window is invisible during whatever gap remains between "window visible to WindowServer" and "WKWebView has painted". `NSAnimationContext.runAnimationGroup` with `duration = 0.15` gives a smooth fade rather than a snap.

The `hasCompletedFirstPaint` flag gates the fade-in so subsequent `didFinish` calls (SPA route changes, `⌘R` reloads) don't re-animate. It's an instance var, so `AppDelegate.openBrowser`'s new `BrowserWindowController` starts fresh. Correct scope.

The `didFailProvisionalNavigation` handler restores `alphaValue = 1` before calling `onNavigationFailed?()` — important: if the very first nav fails, `didFinish` never fires, and the error window transition would otherwise be from an invisible state. Good.

**Fix #37/#45 — Cmd+F find bar.**

The `performKeyEquivalent` rewrite is correctly using `event.modifierFlags.intersection(.deviceIndependentFlagsMask) == .command` rather than the previous `contains(.command)` — this is the AppKit-canonical way to match "exactly Cmd, no other modifiers" and avoids spurious matches on Cmd+Shift+V / Cmd+Opt+V.

The key correctness move for paste-into-find-field: Cmd+V now short-circuits to the web-view paste handler **only when the first responder is not an `NSText`**:

```swift
!(firstResponder is NSText)
```

`NSSearchField` uses an `NSTextField` backing that uses an `NSTextView` as its field editor; `NSTextView` inherits from `NSText`, so `firstResponder is NSText` is true when the find field is focused. The early return lets the default AppKit Cmd+V reach the search field. That's the exact mechanism the user's smoke test checks.

Cmd+Shift+G matching uses `charactersIgnoringModifiers == "G"` (uppercase) — correct, since with Shift held, `charactersIgnoringModifiers` still reports the uppercase form.

Find bar implementation:
- **NSVisualEffectView with `.headerView` material** — comment correctly notes that `.titlebar` is deprecated; `.headerView` is the modern equivalent.
- **Frame sizing**: `autoresizingMask = [.width, .minYMargin]` pins to top and stretches with window resize. `webView.frame.size.height -= barHeight` makes room; `hideFindBar` recomputes from `contentView.bounds.height - statusBarHeight` so any resize that happened while the bar was open is picked up.
- **`cancelOperation(_:)` override** — idiomatic AppKit for Escape while first responder. Closes the bar when visible; delegates to super otherwise. Exactly the smoke-test path.
- **Button glyphs**: `\u{2039}` / `\u{203A}` (‹ ›) — the correct Swift Unicode escape syntax with braces. You learned from v1.3.3.
- **`runFind`**: escapes backslash first, then single-quote — the only correct ordering for JS single-quoted string literals (reverse would double-escape the newly introduced backslashes). `window.find` args `(query, false, backwards, true, false, true, false)` map correctly to `(string, caseSensitive=false, backwards, wrapAround=true, wholeWord=false, searchInFrames=true, showDialog=false)`.
- **Cleanup**: `hideFindBar` called from `windowWillClose` — prevents leaked subviews/first-responder weirdness if the user closes the window while the bar is open.

**Error UX cleanup.**
The `descText` now says "Make sure hermes-webui is running and listening at the URL above. Run: bash start.sh (or: docker compose up -d)" instead of prescribing a specific `cd ~/hermes-webui-public` path. Better — the previous copy assumed a local checkout of a specific repo, which isn't true for users following the docker setup path.

### Smoke tests requested
I cannot run the interactive tests from the build environment (no live user at the keyboard). The four paths were verified statically:

1. **No flash on startup** — the three-layer `#1a1a1a` unification + `alphaValue = 0 → fade` means the window can't show any pre-paint white colour. The only visible state before paint is the animated fade-in from a uniformly-dark background.
2. **Cmd+F opens find bar** — `performKeyEquivalent` matches `⌘ + "f"` → `onFind` → `toggleFindBar` → `showFindBar` → `makeFirstResponder(field)`. All handlers wired.
3. **Escape closes find bar** — `cancelOperation(_:)` override; `findBarVisible` flag gates the behaviour; `hideFindBar` restores webView size and first responder.
4. **Cmd+V pastes into find field** — `!(firstResponder is NSText)` guard in `performKeyEquivalent` cedes Cmd+V to the default system handler when the search field's field editor has focus.

### Non-blocking follow-ups

1. **JS escape in `runFind` doesn't cover newlines.** `NSSearchField` generally strips embedded newlines on Return (because Return fires the action instead of inserting), so the realistic attack surface is near-zero — but if a user ever pastes a multi-line string into the find field, the injected JS would break. A `.replacingOccurrences(of: "\n", with: "\\n")` pair would be cheap insurance.

2. **First-ever nav hangs forever (not fails) = invisible window.** The window stays at `alphaValue = 0` if `didFinish` never fires *and* `didFailProvisionalNavigation` never fires. In practice the network timeout kicks in and surfaces a failure, and the direct-mode preflight catches the common unreachable case before `BrowserWindowController` is even created. Low-probability edge case; a hard-coded 10s-ish timeout that force-shows the window would be belt-and-suspenders.

3. **Infra note (not blocking this PR):** during verification, the `/tmp/swift-mac-test` working tree was out of sync with HEAD (`Package.swift`, `.gitignore`, `CLAUDE.md`, etc. reported as deleted by `git status` while not present on disk). Unrelated to this PR — something between sessions corrupted the working tree. Recovered via `git restore .`. Worth noting for the skill doc if it keeps happening.

Ready to merge and tag v1.4.0.

PR #55, opened 2026-04-23
feat: v1.4.1 — add Edit → Find menu items for Cmd+F discoverability

### LINKED ISSUE #37: feat: find in page (Cmd+F) via WKWebView find API

nesquena-hermes:
## Summary

Follow-up to v1.4.0 (#54): adds Edit → Find → Find… to the menu bar so Cmd+F is discoverable via the standard macOS menu convention.

## What changed

`AppDelegate.setupMenu()` — added a Find submenu under Edit:

```
Edit
  Undo           Cmd+Z
  Redo           Cmd+Shift+Z
  ──────────────
  Cut            Cmd+X
  Copy           Cmd+C
  Paste          Cmd+V
  Select All     Cmd+A
  ──────────────
  Find ▶
    Find…        Cmd+F
    Find Next    Cmd+G
    Find Previous Cmd+Shift+G
```

The three menu actions (`openFind`, `findNext`, `findPrev`) forward directly to the `BrowserWindow`'s existing `onFind`/`onFindNext`/`onFindPrev` closures — no changes to `BrowserWindowController`.

## Pre-merge checklist (for Mac agent)

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout fix/find-menu-item && git pull --rebase

swift build -c release 2>&1
swift test 2>&1
```

Smoke test: launch app, check Edit menu has Find → Find… (Cmd+F), Find Next (Cmd+G), Find Previous (Cmd+Shift+G). Click Find… to confirm it opens the bar.

Closes #37 (menu discoverability aspect — the bar itself shipped in v1.4.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [v1.4.1] — 2026-04-23
+
+### Added
+- **Find submenu in Edit menu** — Edit → Find → Find… (Cmd+F), Find Next (Cmd+G),
+  Find Previous (Cmd+Shift+G). Makes the find feature discoverable via the standard
+  macOS menu convention instead of requiring users to know the shortcut.
+
 ## [v1.4.0] — 2026-04-23
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -313,6 +313,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
             withTitle: "Paste", action: #selector(NSText.paste(_:)), keyEquivalent: "v")
         editMenu.addItem(
             withTitle: "Select All", action: #selector(NSText.selectAll(_:)), keyEquivalent: "a")
+        editMenu.addItem(.separator())
+        // Find submenu (fix #37/#45 — makes Cmd+F discoverable via menu)
+        let findMenuItem = NSMenuItem(title: "Find", action: nil, keyEquivalent: "")
+        let findMenu = NSMenu(title: "Find")
+        findMenu.addItem(
+            withTitle: "Find…", action: #selector(openFind), keyEquivalent: "f")
+        let findNextItem = NSMenuItem(
+            title: "Find Next", action: #selector(findNext), keyEquivalent: "g")
+        findMenu.addItem(findNextItem)
+        let findPrevItem = NSMenuItem(
+            title: "Find Previous", action: #selector(findPrev), keyEquivalent: "G")
+        findPrevItem.keyEquivalentModifierMask = [.command, .shift]
+        findMenu.addItem(findPrevItem)
+        findMenuItem.submenu = findMenu
+        editMenu.addItem(findMenuItem)
         let windowMenuItem = NSMenuItem()
         menuBar.addItem(windowMenuItem)
         let windowMenu = NSMenu(title: "Window")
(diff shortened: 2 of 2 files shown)


nesquena (approved):
Verified on Mac agent at tip `ea6702d`.

### Build/test results
- `swift test` — 22/22 passed

Scope: 37/-0 across 2 files (CHANGELOG + AppDelegate). Pure addition, no existing code touched. Branch is `fix/find-menu-item` rather than `sprint-v1-4-1` — worth noting but fine for a scoped hotfix.

### Code review

**Menu structure follows macOS convention.** Edit menu now has a separator + "Find" item with a submenu containing Find… / Find Next / Find Previous — matching Safari/TextEdit. Parent "Find" item has `action: nil, keyEquivalent: ""` (expands submenu only). Correct.

**Key equivalent encoding:**
- Find… — `keyEquivalent: "f"` (lowercase) → ⌘F
- Find Next — `keyEquivalent: "g"` (lowercase) → ⌘G
- Find Previous — `keyEquivalent: "G"` (uppercase) *plus* explicit `keyEquivalentModifierMask = [.command, .shift]` → ⌘⇧G

The explicit modifier mask on Find Previous is slightly redundant (uppercase "G" alone would make AppKit infer Shift), but belt-and-suspenders doesn't hurt and makes the intent unambiguous. Correct either way.

**Menu ↔ `performKeyEquivalent` interaction.** With the browser frontmost, `BrowserWindow.performKeyEquivalent` still runs first in the event dispatch chain and consumes ⌘F/⌘G/⌘⇧G before the menu gets a chance. The menu items provide *discoverability* (the item shows up in the menu bar with its shortcut) but when the window is key they also provide a redundant second handler. Both paths end at the same `onFind` / `onFindNext` / `onFindPrev` closures → same `toggleFindBar` / `findNext(forward:)` invocation — no divergence, no double-fire.

When the browser isn't frontmost (e.g., Preferences is open), `performKeyEquivalent` on the front window doesn't match, and the menu's key-equivalent dispatch takes over, routing to `openFind` on AppDelegate → `browserWindow?.window as? BrowserWindow).onFind?()`. That's a behaviour *gain*: ⌘F now works from any app window, not just when the browser is focused. Reasonable for a single-window app.

**The forwarding cast** `(browserWindow?.window as? BrowserWindow)?.onFind?()` is a safe downcast: `BrowserWindowController.window` is an `NSWindow` but the instance is always `BrowserWindow` (the custom subclass). When `browserWindow` is nil (splash phase, reconnect), the chain short-circuits — the menu click is a benign no-op rather than a crash. For strict correctness the items could be disabled via `NSMenuValidation` during those phases, but the no-op is harmless.

**Action routing** — the menu items use `#selector(openFind)` with no explicit target, so AppKit sends the action up the responder chain and AppDelegate (NSApp's delegate, chain terminus) picks it up via its `@objc func openFind / findNext / findPrev`. Standard AppKit idiom.

### Non-blocking follow-ups
None this round. Pure UX/discoverability addition with no behaviour change to the actual find logic.

Ready to merge and tag v1.4.1.

PR #58, opened 2026-04-25
feat: v1.5.0 — full-size content view (#57), configurable hotkey (#41), session persistence (#10)

### LINKED ISSUE #10: feat: session persistence — restore conversation across reconnects and app restarts
## Overview

Persist WKWebView session state (cookies, localStorage, conversation state) so that a tunnel reconnect, network blip, or app restart does not lose the current conversation.

## Why

Any reconnect event currently wipes the WKWebView — the user loses their conversation history for that session. For a tool used throughout the day, this is a real reliability gap. Every other chat app preserves state across focus changes and reconnects.

## What needs to persist

1. **Cookies and localStorage** — Hermes WebUI stores session ID and preferences here
2. **Conversation state** — the active chat should survive a reconnect
3. **Scroll position** — optional, but good for long conversations
4. Size of the window - restore the size of the window on relaunch

## Implementation approach

`WKWebViewConfiguration.websiteDataStore` — use `WKWebsiteDataStore.default()` (persistent) instead of `nonPersistent()`. This is already the default for WKWebView, so the main work is ensuring reconnect logic reloads the same URL without resetting the data store.

The reconnect path in `AppDelegate.startTunnel()` currently creates a new `BrowserWindowController` and a new WKWebView on every reconnect. Reusing the existing WebView (or at least preserving the data store) would maintain session state across tunnel drops.

## Note

Updated from original — the previous description framed this around closing/reopening a popover. The same issue applies to the full-window app on reconnect. Still valid and worth fixing.

### LINKED ISSUE #41: feat: configurable global hotkey in Preferences

### LINKED ISSUE #57: feat: full-size content view — extend web content under native title bar for custom web controls
## Summary

@aronprins added a custom title bar to the hermes-webui web UI (part of the chat/transcript redesign in PR #587). The request is for the Mac app to support a **full-size content view** mode — where the web content extends under the native macOS title bar — so the web-based controls in that custom title bar serve as the primary window chrome.

## Current behavior

- `BrowserWindowController` uses `styleMask: [.titled, .closable, .resizable, .miniaturizable]`
- Native macOS title bar sits above the web content (~28px of vertical space lost)
- Web app's custom title bar is doubled up with / hidden behind the native one

## Desired behavior

- Add `.fullSizeContentView` to the window style mask
- Set `titlebarAppearsTransparent = true` so the native title bar area is transparent
- Web content fills the full window height, including the title bar region
- Native traffic light buttons (close/minimize/zoom) remain visible and functional, overlaid on the web content
- Web app's custom title bar area (from @aronprins) naturally aligns with the traffic lights

## Implementation notes

In `BrowserWindowController.init`, change the window init:

```swift
// Before
styleMask: [.titled, .closable, .resizable, .miniaturizable]

// After
styleMask: [.titled, .closable, .resizable, .miniaturizable, .fullSizeContentView]
```

Then after `super.init`:

```swift
window.titlebarAppearsTransparent = true
window.titleVisibility = .hidden
```

The web UI will need `padding-top` on its custom title bar to clear the traffic lights — roughly `28px` on standard displays, `38px` if accounting for the full traffic light zone. `env(titlebar-area-height)` may also work if the Window Controls Overlay API is used.

## Coordination needed

- [ ] Confirm with @aronprins what height/padding their custom title bar uses in the web UI
- [ ] Verify traffic lights render correctly in both windowed and fullscreen modes
- [ ] Test that existing fullscreen persistence (v1.3.0) still works with `.fullSizeContentView`
- [ ] Ensure the SSH status bar (bottom of window) is unaffected

## Related

- #587 — chat transcript redesign (adds the custom web title bar)
- v1.3.0 — fullscreen persistence

nesquena-hermes:
## Summary

Three quality-of-life improvements in one sprint: full-size content view, configurable global hotkey, and session persistence across reconnects.

---

### Full-size content view — closes #57

Web content now extends under the native macOS title bar. The web app's custom `.app-titlebar` element (shipped with the transcript redesign in web UI PR #587, now merged) fills the title bar region, eliminating the doubled native/web header.

**Implementation:**
- `BrowserWindowController.init`: added `.fullSizeContentView` to styleMask + `titlebarAppearsTransparent = true` + `titleVisibility = .hidden` (window title stays set for accessibility, Dock, Window menu)
- `injectTrafficLightWidthVar()`: new helper that measures `window.standardWindowButton(.zoomButton)?.frame.maxX + 12` and injects it as `--traffic-light-width` CSS custom property. Default `80px` at documentStart, refined to exact pixels in `didFinish`.
- `windowDidEnterFullScreen`: resets var to `0px` (traffic lights gone in fullscreen)
- `windowDidExitFullScreen`: restores measured value
- Find bar anchored to `window.contentLayoutRect` (not `contentView.bounds`) to avoid overlapping traffic lights

### Configurable global hotkey — closes #41

The hardcoded Cmd+Shift+H is now user-configurable in Preferences.

**Implementation:**
- New file: `HotkeyRecorderView.swift` — NSView subclass that arms on click, captures the next keyDown with Cmd/Ctrl/Option modifier, displays the combo symbolically (⌘⇧H etc.). Uses `UCKeyTranslate` for keyboard-layout-aware display of printable keys. Bare Delete clears the shortcut; Cmd+Delete can still be bound.
- `PreferencesWindowController`: replaced static label with `HotkeyRecorderView` + hint label. Changes are staged in pending vars and only written to UserDefaults on Save (Cancel discards).
- `AppDelegate.setupGlobalHotkey()`: reads keyCode/modifiers from UserDefaults (defaults registered at launch). `carbonEventHandler` installed once; only `carbonHotKeyRef` is re-registered on Preferences save.
- `AppDelegate.reloadGlobalHotkey()`: unregisters old hotkey ref, calls `setupGlobalHotkey()` with new values, shows NSAlert if registration fails (e.g. Cmd+Space claimed by Spotlight).

### Session persistence on reconnect — closes #10

Tunnel drops, network blips, and manual reconnects no longer destroy the WKWebView. localStorage, cookies, IndexedDB, and scroll position survive.

**Implementation:**
- `BrowserWindowController.reconnectInPlace(targetURL:)`: calls `webView.stopLoading()` then `reload()` (same URL) or `load()` (URL changed). Resets `didReportNavigationFailure` and `isIntentionalClose` defensively.
- `AppDelegate.startTunnel()`: `reuseWindow = browserWindow != nil && browserWindow?.connectionMode == connectionMode`. On reuse: `orderOut(nil)` (hide, don't destroy) + call `reconnectInPlace()` on success. On failure or mode switch: fall back to full window rebuild.
- `urlString` and `connectionMode` on `BrowserWindowController` changed from `private let` to `private(set) var` to support URL/mode updates.

---

## Pre-merge checklist

- [ ] **Mac agent build verification** (required — Swift cannot run on the Linux server):

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout sprint-v1.5.0 && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1

```

Expected: build complete, all tests pass, `bash build.sh` produces Hermes Agent.app.

- [ ] Independent review from nesquena account
- [ ] Verify traffic lights align correctly (not obscured by web content)
- [ ] Verify Preferences hotkey recorder: click to arm, press combo, Save, shortcut fires
- [ ] Verify session persists across Cmd+R reconnect and tunnel drop

---

Closes #57, closes #41, closes #10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Changelog
 
+## [v1.5.0] — 2026-04-25
+
+### Added
+- **Configurable global hotkey** — the hardcoded Cmd+Shift+H shortcut is now user-configurable
+  in Preferences. Click the new recorder field to arm it, press any combo with Cmd/Ctrl/Option,
+  and the shortcut updates immediately. Press Delete while recording to clear (disable) the
+  hotkey. The combo is persisted in UserDefaults and survives app restarts. Closes #41.
+
+### Changed
+- **Full-size content view** — web content now extends under the native macOS title bar using
+  `.fullSizeContentView` + `titlebarAppearsTransparent`. The web app's custom `.app-titlebar`
+  element sits in the title-bar region, eliminating the doubled native/web header. Traffic
+  lights stay visible via a `--traffic-light-width` CSS custom property (default `80px`,
+  refined to the exact measured value after first paint). The variable resets to `0px` in
+  fullscreen and restores on exit. Closes #57.
+
+### Fixed
+- **Session state preserved on reconnect** — tunnel drops, network blips, and manual reconnects
+  no longer destroy the WKWebView. The existing browser window is hidden (orderOut) and reused
+  on reconnect via `reconnectInPlace()`, preserving localStorage, cookies, IndexedDB, and
+  scroll position. The WKWebView is only replaced when the error window takes over on a failed
+  reconnect. Closes #10.
+
 ## [v1.4.1] — 2026-04-23
 
 ### Added
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -38,7 +38,13 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         // Register non-persistent defaults so users upgrading from v1.1.0
         // (where notifications were always on) don't silently lose them.
         // seedDefaultsIfNeeded only persists on first-ever launch.
-        UserDefaults.standard.register(defaults: ["notificationsEnabled": true])
+        UserDefaults.standard.register(defaults: [
+            "notificationsEnabled": true,
+            // Fix #41: default global hotkey = Cmd+Shift+H
+            "globalHotkeyKeyCode": kVK_ANSI_H,
+            "globalHotkeyModifiers": Int(cmdKey | shiftKey),
+            "globalHotkeyEnabled": true,
+        ])
 
         // Initialize Sparkle updater — feed URL comes from SUFeedURL in Info.plist
         updaterController = SPUStandardUpdaterController(
@@ -76,9 +82,18 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let splashSubtitle = connectionMode == "ssh" ? "Establishing SSH tunnel…" : "Connecting…"
         splashWindow = SplashWindowController(title: appTitle, subtitle: splashSubtitle)
         splashWindow.showWindow(nil)
-        browserWindow?.isIntentionalClose = true
-        browserWindow?.close()
-        browserWindow = nil
+        // Fix #10: if the browser window is alive AND the connection mode hasn't changed,
+        // hide it (orderOut) and reuse the WKWebView to preserve session state.
+        // A mode switch (direct↔ssh) must rebuild the window to get the correct status bar.
+        let reuseWindow = browserWindow != nil &&
+            browserWindow?.connectionMode == connectionMode
+        if reuseWindow {
+            browserWindow?.window?.orderOut(nil)
+        } else {
+            browserWindow?.isIntentionalClose = true
+            browserWindow?.close()
+            browserWindow = nil
+        }
         errorWindow?.close()
         errorWindow = nil
         tunnelManager?.stop()
@@ -110,13 +125,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
                 DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                     self.splashWindow.close()
                     if self.tunnelManager.status == .connected {
-                        self.openBrowser(
-                            targetURL: targetURL,
-                            mode: "ssh",
-                            sshHost: host,
-                            localPort: localPort
-                        )
+                        if reuseWindow, let existing = self.browserWindow {
+                            // Fix #10: reuse existing WKWebView for session continuity.
+                            existing.reconnectInPlace(targetURL: targetURL)
+                            existing.window?.makeKeyAndOrderFront(nil)
+                            self.setOfflineBadge(false)
+                        } else {
+                            self.openBrowser(
+                                targetURL: targetURL,
+                                mode: "ssh",
+                                sshHost: host,
+                                localPort: localPort
+                            )
+                        }
                     } else {
+                        self.browserWindow = nil  // abandon the hidden window
                         self.showErrorWindow(targetURL: targetURL, mode: "ssh")
                     }
                 }
@@ -126,13 +149,21 @@ class AppDelegate: NSObject, NSApplicationDelegate {
                 DispatchQueue.main.async {
                     self.splashWindow.close()
                     if reachable {
-                        self.openBrowser(
-                            targetURL: targetURL,
-                            mode: "direct",
-                            sshHost: nil,
-                            localPort: nil
-                        )
+                        if reuseWindow, let existing = self.browserWindow {
+                            // Fix #10: reuse existing WKWebView for session continuity.
+                            existing.reconnectInPlace(targetURL: targetURL)
+                            existing.window?.makeKeyAndOrderFront(nil)
+                            self.setOfflineBadge(false)
+                        } else {
+                            self.openBrowser(
+                                targetURL: targetURL,
+                                mode: "direct",
+                                sshHost: nil,
+                                localPort: nil
+                            )
+                        }
                     } else {
+                        self.browserWindow = nil  // abandon the hidden window
                         self.showErrorWindow(targetURL: targetURL, mode: "direct")
                     }
                 }
@@ -427,6 +458,7 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         if preferencesWindow == nil {
             preferencesWindow = PreferencesWindowController()
             preferencesWindow?.onSave = { [weak self] in
+                self?.reloadGlobalHotkey()  // Fix #41: apply new hotkey from UserDefaults
                 self?.preferencesWindow = nil
                 self?.startTunnel()
             }
@@ -439,7 +471,15 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     // Uses Carbon RegisterEventHotKey — works without Accessibility permission,
     // fires from any app immediately on first launch.
 
+    // MARK: - Global hotkey (configurable, fix #6 + #41)
+
     private func setupGlobalHotkey() {
+        let defaults = UserDefaults.standard
+        // Fix #41: check enabled flag; skip registration when user cleared the shortcut.
+        guard defaults.bool(forKey: "globalHotkeyEnabled") else { return }
+        let keyCode = UInt32(defaults.integer(forKey: "globalHotkeyKeyCode"))
+        let mods    = UInt32(defaults.integer(forKey: "globalHotkeyModifiers"))
+
         var eventSpec = EventTypeSpec(
             eventClass: OSType(kEventClassKeyboard),
             eventKind: UInt32(kEventHotKeyPressed)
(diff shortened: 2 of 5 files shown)


nesquena (approved):
Verified on Mac agent at tip `30349a8`.

### Build/test results
- `swift build -c release` — Build complete (7.10s, 0 warnings)
- `swift test` — 22/22 passed (4 suites: PortValidationTests + SSHArgumentTests + URLValidationTests + LaunchBehaviorTests)

Scope: 446/-48 across 5 files. Matches the PR description: three issues in one commit (#57 full-size content view, #41 configurable hotkey, #10 session persistence).

Verification clean clone — the previous test clone at `/tmp/swift-mac-test` had object-database corruption (`git fsck` reported many missing trees and blobs, `git restore` failed with "unable to read sha1 file"). Re-cloned from origin to recover; same as documented in the skill's gotcha section.

### Code review

**Issue #57 — full-size content view + traffic light width injection.**

Window mask gains `.fullSizeContentView`; `titlebarAppearsTransparent = true`; `titleVisibility = .hidden` (the title text is hidden from the title bar but `window.title` still feeds the Window menu, Dock, Mission Control, and accessibility). Standard recipe for "web app paints under the title bar".

The two-stage CSS variable injection is the right pattern:
1. **At documentStart** — a `WKUserScript` injects `--traffic-light-width: 80px` before any web content runs. This default prevents content rendering under the traffic lights even on first paint.
2. **After didFinish** — `injectTrafficLightWidthVar()` measures `window.standardWindowButton(.zoomButton).frame.maxX + 12` and overrides with the exact pixel value. `.zoomButton` is the rightmost (green) traffic light, and its `frame` is in window-space, so `maxX + 12` is a clean right edge for the web title bar's left margin. Falls back to `80` if the button reference is unavailable.

Fullscreen handling is symmetric: `windowDidEnterFullScreen` resets to `'0px'` (no traffic lights to clear); `windowDidExitFullScreen` re-runs `injectTrafficLightWidthVar()`. The JS interpolates an `Int`, so no escaping concern.

Find bar position fix: previous code anchored to `contentView.bounds.height - barHeight`, which now overlaps the traffic lights because the contentView extends under the title bar. Fixed by anchoring to `window.contentLayoutRect.maxY` (the area below the native title bar). Same change in `hideFindBar` to keep the webView restore consistent. `bar.autoresizingMask = [.width, .minYMargin]` is correct — the bar stays anchored near the top regardless of window resize.

**Issue #41 — configurable global hotkey.**

`HotkeyRecorderView` (new, 215 lines):
- Click-to-arm via `becomeFirstResponder`, drops focus on capture/clear/cancel.
- **Bare-vs-modifier Delete handling** (the Opus catch): `let usableModsForClear = [.command, .control, .option]`; bare Delete (no Cmd/Ctrl/Option) clears, but Cmd+Delete falls through and gets captured as a normal hotkey. Correct semantics.
- Letter-without-modifier guard: `NSSound.beep()` and don't capture — prevents pure-letter global hotkeys that would intercept normal typing.
- **TIS API retain semantics** (the other Opus catch): `TISCopyCurrentKeyboardLayoutInputSource()` is Copy-style → `takeRetainedValue()`. `TISGetInputSourceProperty()` is Get-style → `Unmanaged.fromOpaque(ptr).takeUnretainedValue()`. Both correct, and the comments call out the ownership reasoning.
- `UCKeyTranslate` for printable keys (respects current keyboard layout); symbol table for non-printing keys (arrows, function keys, Return, Tab, Escape, etc.).
- Display order `⌃⌥⇧⌘key` matches macOS standard modifier ordering.
- Uses semantic colors (`controlAccentColor`, `controlBackgroundColor`, `separatorColor`, `labelColor`, `secondaryLabelColor`, `tertiaryLabelColor`) — adapts to dark mode automatically.

`AppDelegate.setupGlobalHotkey` rewrite:
- Reads `keyCode` and `mods` from UserDefaults instead of hardcoding `kVK_ANSI_H` / `cmdKey | shiftKey`.
- Skips registration entirely if `globalHotkeyEnabled == false`.
- **Event handler installed only once** (`if carbonEventHandler == nil`) — important because `reloadGlobalHotkey` is called on every Preferences save, and a stack of identical handlers would fire the callback N times per keypress.
- Defaults registered via `UserDefaults.standard.register(defaults: [...])` so existing users keep `Cmd+Shift+H` after upgrade without overwriting any value they later customize.

`reloadGlobalHotkey`: unregisters the existing hotkey ref (only the ref, not the handler), calls `setupGlobalHotkey` to register a new ref with the new keys, and shows an NSAlert "Shortcut unavailable" if the user wanted it enabled but registration failed (e.g. Cmd+Space taken by Spotlight). The `globalHotkeyEnabled && carbonHotKeyRef == nil` check correctly distinguishes "user disabled it" from "OS rejected it".

**Pending-edit pattern in PreferencesWindowController** (the Cancel-discards Opus catch):
- `pendingHotkeyKeyCode/Modifiers/Enabled` instance vars.
- `onCapture` / `onClear` only set the pending vars — *no* UserDefaults writes.
- `save()` writes the pending values to UserDefaults.
- Cancel just closes — pending values are discarded with the controller. Correct.

**Issue #10 — session persistence on reconnect.**

The reuse decision is gated on connection-mode parity:
```swift
let reuseWindow = browserWindow != nil &&
    browserWindow?.connectionMode == connectionMode
```
A direct↔ssh switch correctly falls through to the rebuild path because the SSH status bar wouldn't be wired up otherwise (this was the Opus mode-switch catch). `connectionMode` and `urlString` were promoted to `private(set) var` so AppDelegate can read them and `reconnectInPlace` can rewrite `urlString`.

`reconnectInPlace` mechanics:
1. `didReportNavigationFailure = false` — reset dedup so a real failure on the new attempt routes to the error window.
2. `isIntentionalClose = false` — defensive, ensures `windowDidExitFullScreen` doesn't no-op.
3. `webView.stopLoading()` — cancels any in-flight provisional load. The cancellation surfaces via `didFailProvisionalNavigation` with `NSURLErrorCancelled`, which is already short-circuited in the existing handler — so no zombie failure event reaches `onNavigationFailed`.
4. Same URL → `webView.reload()`; different URL → updates `urlString` and `webView.load(URLRequest)`.

Window lifecycle on reconnect:
- Reuse path: `orderOut(nil)` (hide, don't destroy) → `reconnectInPlace` → `makeKeyAndOrderFront(nil)` → `setOfflineBadge(false)`.
- Failed reconnect: `browserWindow = nil` to drop the hidden controller (ARC deallocates), then `showErrorWindow`. The "abandon the hidden window" comment makes the intent explicit.

Cookies, localStorage, IndexedDB, and scroll position all persist because the WKWebView and its associated `WKWebsiteDataStore` (the default, persistent one) are unchanged across the reload.

### Non-blocking follow-ups

1. **`reconnectInPlace` pre-reload race window** — there's a small theoretical window between `didReportNavigationFailure = false` and `webView.stopLoading()` where, if the *previous* load's failure callback fired with a non-`NSURLErrorCancelled` code, it would be re-armed and consume the dedup slot before the new reload can use it. Practically WebKit consistently emits `NSURLErrorCancelled` for stopLoading-cancelled navigations, but reordering the reset to *after* stopLoading would close even the theoretical hole.

2. **`injectTrafficLightWidthVar` only fires on `didFinish`** — for SPA route changes within the same document, it doesn't re-run (because didFinish is gated on full-frame loads). If the web app's title-bar component depends on the var being re-asserted across SPA transitions, a `setProperty` re-run isn't needed (the value is on documentElement and persists). Just worth noting.

3. **Working-tree corruption recurrence** — the `/tmp/swift-mac-test` test clone arrived with damaged git objects again. Last sprint: working-tree-only damage, recoverable via `git restore .`. This sprint: object-DB damage, required full re-clone. Two occurrences in two sprints suggests something is touching `.git/objects` between sessions. Worth investigating whether a hook, Spotlight, Time Machine, or something else is at play. Documented the failure mode in the skill's gotcha section.

Ready to merge and tag v1.5.0.

PR #60, opened 2026-04-25
fix: v1.5.1 — hide .app-titlebar-icon overlapping traffic lights

### LINKED ISSUE #59: fix: .app-titlebar-icon overlaps traffic lights after v1.5.0 fullSizeContentView
## Bug

With `.fullSizeContentView` added in v1.5.0, the web app's custom `.app-titlebar` now renders in the native title bar region. The title bar's `.app-titlebar-icon` element (a small SVG Hermes logo) sits at the far left of the bar, placing it right next to — and overlapping — the ✕ (close) traffic light button.

## Fix

Inject a `documentStart` WKUserScript that hides `.app-titlebar-icon` via CSS when running inside the Mac wrapper. The rest of the title bar (title text, sub-label, controls) is unaffected. No web UI change needed.

nesquena-hermes:
## Summary

Hotfix for the icon overlap introduced by v1.5.0's `.fullSizeContentView` change.

## Problem

The web app's `.app-titlebar` has a small SVG Hermes logo (`.app-titlebar-icon`) at the far left. Before v1.5.0 this sat below the native title bar. After v1.5.0 the content view extends under the title bar, so the icon renders right next to the traffic lights — overlapping the ✕ close button.

## Fix

One `WKUserScript` injected at `documentStart` that appends a `<style>` element hiding `.app-titlebar-icon` inside the Mac wrapper. The rest of the title bar (title text, connection badge, controls) is completely unaffected. No change to the web UI is needed.

```swift
let hideIconScript = WKUserScript(
    source: """
        (function() {
            const s = document.createElement('style');
            s.textContent = '.app-titlebar-icon { display: none !important; }';
            (document.head || document.documentElement).appendChild(s);
        })();
        """,
    forMainFrameOnly: true
)
config.userContentController.addUserScript(hideIconScript)
```

## Pre-merge checklist

- [ ] **Mac agent build + visual verification:**

```bash
cd /tmp/swift-mac-test
git fetch origin && git checkout fix-v1.5.1-titlebar-icon && git pull --rebase

swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1
open "Hermes Agent.app"
# Verify: icon is gone from title bar, traffic lights unobstructed, title text still present
```

- [ ] Independent review from nesquena account

Closes #59

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [v1.5.1] — 2026-04-25
+
+### Fixed
+- **Title bar icon no longer overlaps traffic lights** — the web app's `.app-titlebar-icon`
+  SVG logo is hidden when running inside the Mac wrapper via an injected `documentStart`
+  stylesheet. With `.fullSizeContentView` (added in v1.5.0) the icon appeared right next
+  to the close button. The rest of the web title bar is unaffected. Closes #59.
+
 ## [v1.5.0] — 2026-04-25
 
 ### Added
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -265,6 +265,23 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         )
         config.userContentController.addUserScript(trafficLightScript)
 
+        // Fix #59: hide the web app's .app-titlebar-icon (SVG logo) when running in the
+        // Mac wrapper. With .fullSizeContentView the icon sits right next to the traffic
+        // lights and overlaps the close button. The window title and other title bar
+        // controls are unaffected.
+        let hideIconScript = WKUserScript(
+            source: """
+                (function() {
+                    const s = document.createElement('style');
+                    s.textContent = '.app-titlebar-icon { display: none !important; }';
+                    (document.head || document.documentElement).appendChild(s);
+                })();
+                """,
+            injectionTime: .atDocumentStart,
+            forMainFrameOnly: true
+        )
+        config.userContentController.addUserScript(hideIconScript)
+
         contentView.addSubview(webView)
 
         // Only add status bar in SSH mode


nesquena (approved):
Verified on Mac agent at tip `eb47c86`.

### Build/test results
- `swift build -c release` — Build complete (2.43s, 0 warnings)
- `swift test` — 22/22 passed

Scope: 25/-0 across 2 files (CHANGELOG + BrowserWindowController). Pure addition.

### Code review

**Single `WKUserScript` injection, all four moving parts correct:**

1. **Timing — `injectionTime: .atDocumentStart`.** Runs before any web content. The IIFE handles the documentStart case where `document.head` may not exist yet (parser hasn't reached `<head>` yet) via the fallback: `(document.head || document.documentElement).appendChild(s)`. Both placements are valid for `<style>` — browsers process style elements regardless of position.

2. **Scope — `forMainFrameOnly: true`.** The web app's title bar is in the main document, not an iframe. Correct scope; avoids unnecessary work on any embedded iframes.

3. **CSS rule — `'.app-titlebar-icon { display: none !important; }'`.** Class selector targets just the icon element. The `!important` qualifier wins over any author rule giving the icon a layout — necessary because the web app must have *some* `display` rule for the icon to render in the first place. No other elements affected.

4. **String safety.** The CSS rule is a static string literal — no user-controlled data interpolated. No XSS surface.

**Lifecycle considerations:**
- **Reload (incl. v1.5.0 `reconnectInPlace` paths):** WebKit re-runs all registered `WKUserScript`s on every main-frame navigation, so the style element is re-injected on each reload. The hide persists.
- **SPA route changes:** documentStart scripts don't re-fire for in-app navigations, but the injected `<style>` element stays in the DOM unless the SPA replaces `document.head` wholesale (atypical for React/Vue). No re-injection needed.
- **Element timing race:** the SPA renders the icon AFTER our CSS is in the DOM, so `display: none !important` applies as soon as the element is created and matches the selector. No race window.

### Non-blocking follow-ups

1. **Coordination with the web side.** If hermes-webui ever wants the icon visible only on web (and hidden in Mac), they could add a class/data attribute to differentiate environments — at which point this Mac-side hide becomes redundant. Worth a comment in the web repo's title-bar component pointing back at this Swift fix so future web changes don't accidentally reintroduce the overlap by renaming `.app-titlebar-icon`.

Ready to merge and tag v1.5.1.

PR #65, opened 2026-04-28
fix: v1.5.3 — restore window drag after fullSizeContentView change

### LINKED ISSUE #64: After updating Hermes Agent to v1.5.x, the main window can no longer be moved by dragging with the mouse.
## Repro

1. Install or update to Hermes Agent v1.5.0 / v1.5.1 / v1.5.2
2. Launch the app
3. Try to drag the main window by the custom title bar area at the top

## Expected

The window should move when dragging the title bar area, while interactive controls inside explicit non-draggable regions should remain clickable.

## Actual

The window does not move at all.

## Notes

This looks like a regression introduced by the v1.5.0 “full-size content view” change.

Relevant changelog entry:
- `CHANGELOG.md` v1.5.0
- “Full-size content view — web content now extends under the native macOS title bar using .fullSizeContentView + titlebarAppearsTransparent”

Relevant code:
- `Sources/HermesAgent/BrowserWindowController.swift`

The likely problem is that the app moved to a custom web title bar inside `WKWebView`, but native window dragging was not preserved after switching to:

- `.fullSizeContentView`
- `titlebarAppearsTransparent = true`
- `titleVisibility = .hidden`

The web UI already appears to use `-webkit-app-region: drag` / `no-drag` semantics in its custom title bar, but `WKWebView` does not automatically map that to native `NSWindow` dragging.

## Additional context

I confirmed this is not just stale app state:
- existing installed app version was `1.5.2`
- issue reproduces after relaunch
- this did not happen before the full-size title bar change

nesquena-hermes:
## Problem

After the v1.5.0 `.fullSizeContentView` + `titlebarAppearsTransparent` change, the main window can no longer be moved by dragging the title bar area (issue #64).

**Root cause:** `WKWebView` now fills the entire `contentView` including the transparent native title bar strip, and intercepts all mouse events. The `-webkit-app-region: drag` CSS in the web UI has no effect on `NSWindow` dragging — that's a browser-tab chrome feature, not a macOS window management API.

## Fix

Add a private `TitleBarDragView: NSView` subclass — a fully transparent overlay view positioned over the title bar zone (`y = bounds.height - 38` to `bounds.height`, matching `.app-titlebar { height: 38px }` in the web UI). It:

- Calls `window?.performDrag(with: event)` on single `mouseDown`, restoring native drag
- Handles double-click by reading the system `AppleActionOnDoubleClick` preference (Maximize / Minimize / None), so the standard "double-click title bar to zoom/restore" gesture still works
- Uses `acceptsFirstMouse(for:) → true` so the first click on an inactive window both activates and drags

The view is added to `contentView` after `WKWebView`, so it sits on top in z-order and wins hit-testing in the title bar zone. Traffic lights live in `NSThemeFrame` above `contentView` and are completely unaffected.

## Testing

Please run on the Mac agent:

```bash
cd /tmp/swift-mac-test
git clone https://github.com/hermes-webui/hermes-swift-mac . 2>/dev/null || \
  (cd /tmp && rm -rf swift-mac-test && git clone https://github.com/hermes-webui/hermes-swift-mac swift-mac-test && cd swift-mac-test)
git fetch origin && git checkout fix/window-drag-regression && git pull --rebase

swift package resolve
swift build -c release 2>&1
swift test 2>&1
bash build.sh 2>&1
```

**Manual verify after `bash build.sh`:**
1. Launch `Hermes Agent.app`
2. Drag the title bar area — window should move ✓
3. Double-click the title bar — window should zoom/restore ✓
4. Traffic lights (close/minimize/zoom buttons) should all still respond ✓
5. Interactive controls in the title bar (e.g., hamburger on mobile viewport) should still receive clicks ✓

Closes #64

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [v1.5.3] — 2026-04-28
+### Fixed
+- **Window drag regression** — after the v1.5.0 `.fullSizeContentView` change, the main window could no longer be moved by dragging the title bar. `WKWebView` covers the full content area including the transparent native title bar strip and intercepts all mouse events; `-webkit-app-region: drag` in the web page's CSS has no effect on `NSWindow` dragging. Fixed by adding a thin transparent `TitleBarDragView` overlay positioned over the title bar zone (height 38 px, matching `.app-titlebar` in the web UI) that calls `window.performDrag(with:)` on `mouseDown`. The view is fully transparent and only captures hits within its own bounds. Traffic lights live in `NSThemeFrame` above `contentView` and are unaffected. (fixes #64)
+
 ## [v1.5.2] — 2026-04-25
 
 ### Fixed
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -45,6 +45,36 @@ private class HermesWebView: WKWebView {
     override func acceptsFirstMouse(for event: NSEvent?) -> Bool { true }
 }
 
+// Fix #64: transparent drag view that sits atop the WKWebView in the title-bar zone.
+// With .fullSizeContentView + titlebarAppearsTransparent, WKWebView covers the native
+// title bar strip and intercepts all mouse events — killing native window drag.
+// -webkit-app-region: drag in the web page's CSS has no effect on NSWindow dragging.
+// This overlay calls window.performDrag(with:) on mouseDown in the title-bar strip,
+// restoring the expected drag-to-move behaviour. The view is fully transparent
+// (no layer, no drawing) so it has no visual impact. Traffic lights live in
+// NSThemeFrame above contentView and are unaffected.
+private class TitleBarDragView: NSView {
+    override func acceptsFirstMouse(for event: NSEvent?) -> Bool { true }
+
+    override func mouseDown(with event: NSEvent) {
+        // Double-click: honour the system "Double-click a window's title bar to" preference.
+        if event.clickCount == 2 {
+            let action = UserDefaults.standard.string(forKey: "AppleActionOnDoubleClick") ?? "Maximize"
+            switch action {
+            case "Minimize": window?.miniaturize(nil)
+            case "Maximize": window?.performZoom(nil)
+            default: break  // "None"
+            }
+            return
+        }
+        // Single click: pass to the window's native drag-to-move handler.
+        window?.performDrag(with: event)
+    }
+    // No hitTest override — default NSView.hitTest is correct (point is in superview coords,
+    // default returns self when point is in frame, nil otherwise).
+    // No isFlipped override — the view has no subviews or drawing; isFlipped is irrelevant.
+}
+
 class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegate, WKNavigationDelegate, WKScriptMessageHandler {
 
     private var webView: HermesWebView!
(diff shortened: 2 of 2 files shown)


nesquena (approved):
Verified on Mac agent at tip `308fb25`.

### Build/test results
- `swift build -c release` — Build complete (3.02s, 0 warnings)
- `swift test` — 22/22 passed

Scope: 51/-0 across 2 files (CHANGELOG + BrowserWindowController). Pure addition.

### Code review

**The diagnosis is correct and worth restating clearly:** `-webkit-app-region: drag` is browser-tab-chrome territory (Electron, Chrome packaged apps). It has no effect on `NSWindow` dragging. With `.fullSizeContentView` + `titlebarAppearsTransparent`, `WKWebView` covers the whole `contentView` including the title-bar strip and intercepts every mouse event before AppKit's title bar drag logic ever sees it. The overlay-view approach is the standard AppKit answer.

**Geometry & z-order:**
- Frame is `NSRect(x: 0, y: bounds.height - 38, width: bounds.width, height: 38)` — the top 38pt of the contentView in AppKit's bottom-up coords. Matches the web UI's 38pt `.app-titlebar` height per the commit message.
- `autoresizingMask = [.width, .minYMargin]` — width stretches with the window; the minY-margin is flexible while height stays fixed at 38, which keeps the view anchored to the top edge during resize. Correct AppKit autoresizing for a top-anchored strip.
- Added via `contentView.addSubview(dragView)` *after* `webView`, so it sits on top in z-order. NSView hit testing walks subviews in reverse order (last-added wins), so the overlay intercepts mouse events in its frame before WKWebView gets them. Below the 38pt strip, WKWebView's hit tests succeed normally — web content, find bar, etc. all unaffected.

**Traffic light interaction:** correctly noted in the comment — the close/minimize/zoom buttons live in `NSThemeFrame` *above* `contentView`, not inside it. The overlay can't intercept clicks on them. So Cmd+W works, native traffic-light click tracking works, and only the empty title-bar zone (the area where the web's custom title bar is painted under transparent native chrome) is captured by the overlay.

**Mouse handling correctness:**
- `acceptsFirstMouse(_:)` returns `true` — clicks on a backgrounded window's title strip both raise the window AND start a drag in one motion (matches AppKit's native title bar behaviour).
- Single click → `window?.performDrag(with: event)`. `performDrag` enters a modal tracking loop only if the user actually moves the mouse; on a pure click without drag it returns immediately, so it doesn't swallow the click.
- Double click → reads `AppleActionOnDoubleClick` from `UserDefaults` and dispatches `miniaturize` / `performZoom` / no-op for `"Minimize"` / `"Maximize"` / `"None"`. Default fallback is `"Maximize"`, which matches the macOS factory default. The code returns *before* calling `performDrag` in the double-click case, so the second click doesn't accidentally start a drag tracking loop.
- Single-vs-double race: AppKit dispatches sequential `mouseDown` events with incrementing `clickCount`. The first click in a double-click sequence has `clickCount == 1` and falls through to `performDrag` (immediate return on no-drag), then the second click arrives with `clickCount == 2` and runs the configured action. Both paths work.

**Find bar non-interaction:** the find bar anchors to `window.contentLayoutRect.maxY - barHeight`, which is below the native title bar strip. The drag overlay (top 38pt of contentView) and the find bar live in separate y-ranges; no z-order or hit-test overlap when the find bar is shown.

**Lifecycle:** `titleBarDragView` is held as an instance property. `contentView.addSubview` adds a strong reference too. ARC handles cleanup when the controller goes away. The v1.5.0 `reconnectInPlace` path doesn't touch the overlay (it operates on `webView`), so the drag behaviour persists across reconnects. Mode switches that build a fresh `BrowserWindowController` re-create the overlay in `buildUI`. Good.

**No drawing, no isFlipped, no hitTest override** — the comments correctly call out why each is unnecessary. The view is a pure event-capture surface.

### Non-blocking follow-ups

1. **Hardcoded 38pt** matching the web's `.app-titlebar` height. If the web team ever changes that height (e.g. to make it more macOS-native at ~28pt), this Swift constant won't auto-track and the drag-zone height would mismatch the visual title bar height. Worth a comment in the web repo's title-bar component pointing at `BrowserWindowController.swift` so a height change there reminds someone to update here. Same class of cross-repo coupling as the `.app-titlebar-icon` selector in v1.5.1/v1.5.2.

Ready to merge and tag v1.5.3.

PR #66, opened 2026-04-30
fix: intercept WKWebView downloads with NSSavePanel instead of render…

spinuplabs:
…ing raw content

WKWebView renders responses inline by default — clicking a download link loaded raw data into the window instead of prompting for a save location.

Intercept downloads in decidePolicyFor navigationResponse by checking the Content-Disposition: attachment header and canShowMIMEType. Hand off to WKDownloadDelegate, which presents NSSavePanel pre-filled with the server-suggested filename. Download failures surface as a sheet alert.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [v1.5.4] — 2026-05-02
+
+### Fixed
+- **Download links now save instead of rendering raw content** — clicking a download link in the web UI loaded the raw response data into the WKWebView window instead of prompting for a save location. The `decidePolicyFor navigationResponse` handler now intercepts responses whose `Content-Disposition` header begins with `attachment` (case-insensitive) **or** whose MIME type WebKit can't render, and hands them off to `WKDownload`. The new `WKDownloadDelegate` extension presents an `NSSavePanel` pre-filled with the server's suggested filename, runs as a sheet on the main window, and surfaces failures via a sheet alert. Closes the gap where attachments from the chat UI (file exports, downloads from previous sessions, etc.) had no way to be saved. (#66, thanks @redsparklabs)
+
 ## [v1.5.3] — 2026-04-28
 ### Fixed
 - **Window drag regression** — after the v1.5.0 `.fullSizeContentView` change, the main window could no longer be moved by dragging the title bar. `WKWebView` covers the full content area including the transparent native title bar strip and intercepts all mouse events; `-webkit-app-region: drag` in the web page's CSS has no effect on `NSWindow` dragging. Fixed by adding a thin transparent `TitleBarDragView` overlay positioned over the title bar zone (height 38 px, matching `.app-titlebar` in the web UI) that calls `window.performDrag(with:)` on `mouseDown`. The view is fully transparent and only captures hits within its own bounds. Traffic lights live in `NSThemeFrame` above `contentView` and are unaffected. (fixes #64)
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -659,16 +659,25 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         decidePolicyFor navigationResponse: WKNavigationResponse,
         decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void
     ) {
-        if let httpResponse = navigationResponse.response as? HTTPURLResponse,
-            httpResponse.statusCode >= 500
-        {
-            decisionHandler(.cancel)
-            guard !didReportNavigationFailure else { return }
-            didReportNavigationFailure = true
-            onNavigationFailed?()
-        } else {
-            decisionHandler(.allow)
+        if let httpResponse = navigationResponse.response as? HTTPURLResponse {
+            if httpResponse.statusCode >= 500 {
+                decisionHandler(.cancel)
+                guard !didReportNavigationFailure else { return }
+                didReportNavigationFailure = true
+                onNavigationFailed?()
+                return
+            }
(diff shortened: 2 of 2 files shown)


nesquena-hermes (approved):
## Review — looks good, ship after Mac build verification

Thanks @redsparklabs — this is a clean, well-targeted fix. Reviewed and approved on my side.

### What's correct

- Intercept happens at the right hook (`decidePolicyFor navigationResponse:`), and the `WKDownload` handoff via `didBecome download:` is the standard pattern for macOS 11.3+
- Trigger condition is good: `Content-Disposition: attachment` (case-insensitive `hasPrefix` covers `attachment; filename=…`) **OR** `!canShowMIMEType` — the second clause is important because plenty of servers send `application/zip` / `application/octet-stream` without any disposition header
- The 5xx branch correctly `return`s before falling through, so error pages are never mistakenly treated as downloads
- `NSSavePanel` is pre-filled with the server's `suggestedFilename` and runs as a sheet on the main window — matches Safari's behaviour
- Failure path shows a sheet alert with the error description, dispatched on the main queue with a `[weak self]` capture
- Deployment target is `.macOS(.v12)`; all APIs used are 11.3+ so no `@available` guards needed
- CI green, mergeable, no conflicts with main

### One small follow-up

I pushed a tiny cleanup commit (`ad6a5e1`, -2 lines) removing an orphan `// MARK: - Download (WKDownloadDelegate wiring)` comment that landed in the middle of the class — the actual implementation sits in the `WKDownloadDelegate` extension at the bottom of the file, so the standalone marker was confusing. Hope that's OK; happy to revert if you'd rather keep it.

### Pre-merge

Will run `swift build -c release` + `swift test` + `bash build.sh` on the Mac agent to confirm the universal binary builds clean, then merge and ship as v1.5.4. Will credit you in CHANGELOG and the GitHub release notes.

nesquena (approved):
Verified on Mac agent at tip `ad6a5e1`.

### Build/test results
- `swift build -c release` — Build complete (7.67s after fresh clone, 0 warnings)
- `swift test` — 22/22 passed

Scope: 51/-9 across 1 file (BrowserWindowController). Two commits — `ebec6f1` (download interception + delegate) and `ad6a5e1` (cosmetic: drop an orphan `MARK` header pointing at the extension at the bottom of the file). Test clone was missing on disk again — re-cloned, same recurring infra issue noted last sprint.

### Code review

**`decidePolicyFor navigationResponse` rewrite — clean.**

The 5xx-error path is preserved verbatim (cancel + dedup-guarded `onNavigationFailed`). Two new download triggers added, both correct:

1. `Content-Disposition: attachment` (case-insensitive prefix match) — handles standard downloads like `attachment; filename="foo.zip"`. The `hasPrefix("attachment")` correctly ignores any params/filename suffix; `lowercased()` handles `Attachment`/`ATTACHMENT` variants.
2. `!navigationResponse.canShowMIMEType` — WebKit can't render the response (e.g. `application/octet-stream`, `application/zip`, etc.), so promote to download instead of rendering raw bytes inline. This covers servers that omit `Content-Disposition` but serve binary types.

Decision priority is right: 5xx wins over download (a 503 Service Unavailable with `Content-Disposition: attachment` is still a server failure and should route to the error window). Renderable types with `Content-Disposition: attachment` correctly download instead of render — explicit user-intent header wins over WebKit's "I can show this" capability.

**`navigationResponse: didBecome download` wiring — minimal and correct.**

Just sets `download.delegate = self` so the `WKDownloadDelegate` extension below picks up the lifecycle calls.

**`WKDownloadDelegate` extension:**
- `decideDestinationUsing` — `NSSavePanel` pre-filled with the server's `suggestedFilename`. `beginSheetModal(for: win)` attaches as a sheet (non-blocking, scoped to the window). On `OK` returns `panel.url`; on cancel returns `nil` (which `WKDownload` interprets as "abort"). The `guard let win = window else { completionHandler(nil); return }` keeps the contract honored even if the window is gone (no leaked completion handler).
- `downloadDidFinish` — empty no-op. Acceptable; macOS users can locate the file via the saved path. A "reveal in Finder" notification could be a future enhancement but isn't required.
- `didFailWithError` — `DispatchQueue.main.async` wraps the alert (correct: WKDownload callbacks may fire off-main), `[weak self]` for the closure, and the alert is presented as a sheet on the window. `resumeData` is intentionally dropped — no resume UI yet, which is fine for a first-pass download fix.

**Concurrent downloads:** if the user clicks two download links rapidly, both `decideDestinationUsing` callbacks fire and each tries to attach a `NSSavePanel` sheet. AppKit serializes sheets per-window automatically — the second sheet displays after the first is dismissed. WKDownload completion handlers stay paired correctly. No race.

**No retain cycle concern:** `WKDownload` holds `delegate = self` (BrowserWindowController) for the lifetime of the download. The download completes (success/fail) and releases the delegate ref; no leak.

### Non-blocking follow-ups

1. **Resume on retry.** `didFailWithError` receives `resumeData` but the current alert just reports the error. A "Retry" button on the alert that called `WKDownload.resumeDownload(fromResumeData:)` would be a clean small follow-up — but only matters if the network is flaky enough that real-world downloads fail mid-stream. Not blocking.

2. **Non-200 success codes.** The 5xx check is `>= 500`; nothing handles 4xx. WebKit will happily render a 404 HTML body inline (which is fine) or trigger download for a 4xx with `Content-Disposition: attachment`. Probably correct as-is — many servers serve 401/403 challenge pages with rich HTML, and intercepting those would break auth flows. Worth a glance from someone with hermes-webui's auth model in mind, but no obvious bug.

3. **The polish commit `ad6a5e1`** is a bona fide cleanup — the `// MARK: - Download (WKDownloadDelegate wiring)` header inside the main class body had no implementation under it (the impl lives in the bottom extension), so it would have shown up as an empty section in Xcode's symbol nav. Removing it is right.

Ready to merge.

PR #67, opened 2026-05-02
feat: v1.6.0 — multi-window + native tabs (#42)

### LINKED ISSUE #42: feat: multi-window or native tab bar support

nesquena-hermes:
## Multi-window + native tabs (#42)

Adds support for multiple independent Hermes sessions in one app — separate windows or native macOS tabs, depending on the user's "Prefer Tabs" system preference.

### What's in here

- **`Cmd+N` → New Window**, **`Cmd+T` → New Tab** — both call `newBrowserWindow()` on AppDelegate; AppKit decides window-vs-tab based on the user's preference
- **Each window owns its own `WKWebView`** — server-side sessions are independent, so chat history, scroll position, in-flight streams, cookies, and localStorage are preserved per window
- **AppKit's native tab system fills in the rest** — `Show Tab Bar`, `Show All Tabs`, `Move Tab to New Window`, `Merge All Windows`, the tab-bar plus button, drag-out-to-window, drag-back-to-merge — all wired automatically once `tabbingMode = .preferred` + `tabbingIdentifier` is set
- **Reconnect fans out** — SSH tunnel restore, `NWPathMonitor` recovery, manual retry — every WKWebView gets `reconnectInPlace(targetURL:)` so no session is lost
- **Sensible window stacking** — first window restores its persisted frame; subsequent windows cascade from the front-most so they're visibly stacked (autosave only on the primary)
- **Menu actions target the front-most window** — Find, Reload, Zoom, Find Next/Previous all operate on `keyBrowserWindow` (resolves via `NSApp.keyWindow.windowController` → falls back to the most recent)

### Architecture

`AppDelegate.browserWindow: BrowserWindowController?` becomes `browserWindows: [BrowserWindowController]`. Every callsite that touched the singular got walked: `startTunnel` reuse logic, SSH/direct reconnect paths, `showErrorWindow`, the global hotkey handler, `applicationShouldHandleReopen`, the menu action selectors, the navigation-failed handler. `onWindowWillClose` callback prunes the array on every close (user, drag-out, programmatic).

### Design decisions worth flagging

1. **One `tabbingIdentifier` for all windows** — `"ai.get-hermes.HermesAgent.main"`. Lets the user merge any two Hermes windows into a tab group via Window → Merge All Windows. If we wanted to stop that we'd need per-window IDs but I can't think of why a user would.
2. **NavigationFailed is per-window, not global** — if one window's WKWebView fails to navigate (e.g. session URL got invalidated server-side), only that window goes away; the others stay live. Only when the LAST window dies do we escalate to the error screen.
3. **Mode-switch teardown is global** — switching between SSH tunnel and direct mode tears down all browser windows and rebuilds. Reusing was tempting but the per-window status bar UI is mode-specific (built once in `buildUI`).
4. **`applicationShouldTerminateAfterLastWindowClosed` stays `false`** — closing all windows leaves the app in the Dock; Cmd+N from there falls through to `startTunnel()` if there's no live connection. Existing behaviour unchanged.

### Testing

- Build verified on the Mac agent: `swift build -c release`, `swift test`, `bash build.sh` all green
- Manual scenarios I exercised (or want the reviewer to exercise):
  - Cmd+N twice → 3 windows cascade visibly, all show the same target URL, all are independent
  - With "Prefer Tabs: Always" set: Cmd+N → tab; tab-bar plus button → tab; drag tab out → independent window; close the dragged-out window → other tabs unaffected
  - With "Prefer Tabs: Never": Cmd+N → separate window; Window → Merge All Windows → both become tabs in one window
  - Open 3 windows, kill the SSH tunnel — all 3 windows show "disconnected" in their status bar; bring tunnel back via Retry → all 3 reload in place, no session loss
  - Network cable yank + reattach → `NWPathMonitor` triggers reconnect; all 3 windows recover
  - Cmd+W last window → app stays in Dock; click Dock → Cmd+N restores via startTunnel
  - In SSH mode with tunnel disconnected: Cmd+N is a no-op (won't create instant-fail windows)

### Opus pre-commit follow-ups

Ran the Swift through the Opus mentor advisor before pushing (we have no Swift compiler on Linux). Caught four real correctness bugs that all-targeted-tests would have missed:

1. **Retain cycle in `onNavigationFailed`** — closure stored on `browser.onNavigationFailed` strongly captured `browser` itself, leaking the BrowserWindowController and its WKWebView for the lifetime of the app on every navigation failure. Fixed with `[weak browser]` capture list.
2. **`windowShouldClose` returning `false` unconditionally** — was the right call with one window (Cmd+W → hide → keep app in Dock). With multi-window, closing a tab via `orderOut(nil)` does NOT fire `windowWillClose`, so `onWindowWillClose` never fires, so AppDelegate never prunes the closed window from `browserWindows`. Closed tabs would phantom in the array; menu actions could resurrect them. Fixed by checking `liveCount <= 1` — only the LAST window hides; all others close normally so the cleanup chain runs.
3. **`windowDidExitFullScreen` clobbered the saved fullscreen pref** — exiting fullscreen on one window while others remained fullscreen would set `windowWasFullScreen = false`, so on next launch no window would restore. Fixed with a `browserWindows.contains { other in other !== self && fullscreen }` guard.
4. **`alphaValue = 0` fade-in for non-first tabbed windows** — when a new tab joins an existing tab group, AppKit selects the new tab; alpha=0 hides the content area while the tab bar already shows the tab as active. Reads as a flash. Fixed by gating the fade-in to `isFirstWindow` only.

Engineering-notes paragraph in the v1.6.0 CHANGELOG entry documents these for future me.

### Closes

Closes #42.

### CHANGELOG

Entry added for v1.6.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [v1.6.0] — 2026-05-02
+
+### Added
+- **Multi-window and native tabs (#42)** — open multiple independent Hermes sessions from the same app. `Cmd+N` opens a new window; `Cmd+T` does the same and lands as a tab when the user's "Prefer Tabs" system preference is set to Always (or In Full Screen). Each window owns its own `WKWebView` so server-side sessions stay independent — chat history, scroll position, in-flight streams, cookies, and localStorage are preserved per window. AppKit's native tabbing system fills in `Show Tab Bar`, `Show All Tabs`, `Move Tab to New Window`, and `Merge All Windows` automatically in the Window menu, plus the tab-bar plus button (wired through `newWindowForTab`). The frontmost browser window receives all menu actions (Find, Zoom, Reload); the global hotkey (default `Cmd+Shift+H`) and Dock-icon click both surface the most-recently-active window. Reconnect logic (SSH tunnel restore, network recovery via `NWPathMonitor`, manual retry) fans out across every open window — every WKWebView reloads in place, no session loss. Window frame autosave applies only to the first window of a session; subsequent windows cascade from the front-most window's frame so they're visibly stacked. Closes #42.
+
+  Engineering notes: Opus pre-commit advisor caught four real correctness bugs before this shipped — (1) `onNavigationFailed` retain cycle leaking the BrowserWindowController + WKWebView for every nav failure, (2) `windowShouldClose` returning `false` unconditionally caused closed tabs to phantom in `browserWindows` because `windowWillClose` doesn't fire on `orderOut`, (3) `windowDidExitFullScreen` clobbered `windowWasFullScreen` even when other windows remained fullscreen, (4) `alphaValue=0` fade-in created a visible flash on tabs joining an existing tab group. All four fixed inline before push.
+
 ## [v1.5.4] — 2026-05-02
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -16,7 +16,29 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
     var tunnelManager: TunnelManager!
     var splashWindow: SplashWindowController!
-    var browserWindow: BrowserWindowController?
+    /// Tracks the next cascade origin for non-first windows. Set by
+    /// `cascadeTopLeft(from:)` in openBrowser so rapid Cmd+N produces a clean diagonal
+    /// stack rather than every new window cascading from the same key-window position.
+    /// Reset to nil whenever the array empties (no live windows → next opening is "first").
+    private var nextCascadePoint: NSPoint?
+    /// Active browser windows. Multiple windows enable concurrent sessions;
+    /// AppKit's window-tabbing system groups them when the user prefers tabs.
+    /// Order: most-recently-opened last. The "key" window for menu actions is
+    /// derived from NSApp.keyWindow when present, falling back to the array tail.
+    var browserWindows: [BrowserWindowController] = []
+    /// The browser window targeted by menu actions (Find, Zoom, Reload, etc).
+    /// Returns the front-most browser window if focused, otherwise the most
+    /// recently opened one. Returns nil only when no browser window is open
+    /// (e.g. error state).
+    var keyBrowserWindow: BrowserWindowController? {
+        if let w = NSApp.keyWindow?.windowController as? BrowserWindowController {
+            return w
+        }
+        if let w = NSApp.mainWindow?.windowController as? BrowserWindowController {
+            return w
+        }
+        return browserWindows.last
+    }
     var errorWindow: ErrorWindowController?
     var preferencesWindow: PreferencesWindowController?
     var updaterController: SPUStandardUpdaterController!
@@ -82,17 +104,22 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         let splashSubtitle = connectionMode == "ssh" ? "Establishing SSH tunnel…" : "Connecting…"
         splashWindow = SplashWindowController(title: appTitle, subtitle: splashSubtitle)
         splashWindow.showWindow(nil)
-        // Fix #10: if the browser window is alive AND the connection mode hasn't changed,
-        // hide it (orderOut) and reuse the WKWebView to preserve session state.
-        // A mode switch (direct↔ssh) must rebuild the window to get the correct status bar.
-        let reuseWindow = browserWindow != nil &&
-            browserWindow?.connectionMode == connectionMode
-        if reuseWindow {
-            browserWindow?.window?.orderOut(nil)
+        // Fix #10 (multi-window): reuse all browser windows in place when the connection
+        // mode hasn't changed — preserves WKWebView state (cookies, scroll, in-flight chat)
+        // for every open session. A mode switch (direct↔ssh) must rebuild every window so
+        // the status-bar UI matches the new mode. We snapshot the array because openBrowser
+        // and showErrorWindow can mutate browserWindows during the reconnect flow.
+        let reuseWindows = !browserWindows.isEmpty &&
+            browserWindows.allSatisfy { $0.connectionMode == connectionMode }
+        if reuseWindows {
+            browserWindows.forEach { $0.window?.orderOut(nil) }
         } else {
-            browserWindow?.isIntentionalClose = true
-            browserWindow?.close()
-            browserWindow = nil
+            browserWindows.forEach { win in
+                win.isIntentionalClose = true
+                win.close()
+            }
+            browserWindows.removeAll()
+            nextCascadePoint = nil
         }
         errorWindow?.close()
         errorWindow = nil
@@ -118,17 +145,19 @@ class AppDelegate: NSObject, NSApplicationDelegate {
 
             tunnelManager.onStatusChange = { [weak self] status in
                 guard let self = self else { return }
-                self.browserWindow?.updateStatus(status, host: host, port: localPort)
+                self.browserWindows.forEach { $0.updateStatus(status, host: host, port: localPort) }
             }
 
             tunnelManager.start {
                 DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
                     self.splashWindow.close()
                     if self.tunnelManager.status == .connected {
-                        if reuseWindow, let existing = self.browserWindow {
-                            // Fix #10: reuse existing WKWebView for session continuity.
-                            existing.reconnectInPlace(targetURL: targetURL)
-                            existing.window?.makeKeyAndOrderFront(nil)
+                        if reuseWindows && !self.browserWindows.isEmpty {
+                            // Fix #10: reuse every existing WKWebView for session continuity.
+                            for win in self.browserWindows {
+                                win.reconnectInPlace(targetURL: targetURL)
+                            }
+                            self.browserWindows.last?.window?.makeKeyAndOrderFront(nil)
                             self.setOfflineBadge(false)
                         } else {
                             self.openBrowser(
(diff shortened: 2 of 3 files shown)


nesquena (approved):
Verified on Mac agent at tip `95f765e`.

### Build/test results
- `swift build -c release` — Build complete (3.04s, 0 warnings)
- `swift test` — 22/22 passed
- `test -d "Hermes Agent.app"` → BUILD OK

Scope: 291/-64 across 3 files (CHANGELOG + AppDelegate + BrowserWindowController). Single feature commit + 4 pre-commit fixes.

### AppKit API verification (vs developer.apple.com)

I cross-checked every new AppKit/NSWindow API call against Apple's official documentation:

- **`NSWindow.tabbingMode = .preferred`** — `NSWindow.TabbingMode` enum, three cases (`.automatic` / `.disallowed` / `.preferred`), `.preferred` means "explicitly prefers to tab together with windows sharing the same tabbing identifier." Available macOS 10.12+. Correct usage.
- **`NSWindow.tabbingIdentifier = "..."`** — typealias for `String` via `NSWindow.TabbingIdentifier`. No documented constraints on the string value. Correct usage.
- **`@objc func newWindowForTab(_ sender: Any?)`** — actually declared on `NSResponder`, not `NSWindow`. AppKit looks up the responder chain when the tab-bar "+" is clicked, including the `NSApplication` delegate. The implementation on `AppDelegate` is correctly picked up.
- **`NSWindow.cascadeTopLeft(from:) -> NSPoint`** — moves the receiver to the given top-left and returns a seed point for the next call. The PR's `nextCascadePoint = win.cascadeTopLeft(from: from)` capture-and-feed pattern matches Apple's documented usage. The preceding `win.setFrame(anchor.frame, display: false)` is required because `cascadeTopLeft` only repositions, not resizes.
- **`NSWindow.addTabbedWindow(_:ordered:)`** — *not* used here. The PR relies on `.preferred` + matching `tabbingIdentifier` to auto-join tab groups. Apple's docs say `.preferred` "explicitly prefers to tab" with same-identifier windows, which is the documented mechanism for the auto-join behaviour the PR wants.

### One inaccuracy in a comment (not blocking)

`AppDelegate.setupMenu` has a comment claiming `NSApp.windowsMenu = windowMenu` "makes AppKit auto-populate it … AND auto-inject 'Show Tab Bar', 'Show All Tabs', 'Move Tab to New Window', and 'Merge All Windows' items for windows whose tabbingMode is .preferred." Per Apple's docs, those tab-related menu items appear because the windows have `tabbingMode != .disallowed`, not because of the `windowsMenu` designation. The functional outcome the author observed is correct (the items do appear) but the cause-and-effect in the comment is wrong — `windowsMenu` only controls the standard "Bring All to Front" + per-window list. Worth a one-line comment edit in a follow-up.

### The four Opus pre-commit fixes — all correctly applied

**1. Retain cycle in `onNavigationFailed`.** `[weak self, weak browser]` capture list, with a `guard let self = self, let browser = browser else { return }` unwrap. Verified the closure is stored on `browser.onNavigationFailed`, so a strong capture of `browser` would create a `controller → closure → controller` cycle leaking the WKWebView. The weak capture breaks it correctly.

**2. `windowShouldClose` `liveCount <= 1` gate.** At the moment `windowShouldClose` fires, the closing window is still in `browserWindows` (it would only be removed via `windowWillClose`, which only fires if this method returns `true`). So `liveCount <= 1` correctly means "this is the last live window." For all other cases, returns `true` → AppKit closes normally → `windowWillClose` fires → `onWindowWillClose` callback prunes the array. Without this fix, closed tabs would phantom in `browserWindows` and menu actions could resurrect dead controllers.

**3. `windowDidExitFullScreen` guard against clobbering peers.** The `appDelegate?.browserWindows.contains { other in other !== self && (other.window?.styleMask.contains(.fullScreen) ?? false) }` predicate correctly excludes the current window (by identity) and only returns true if any *other* window is still fullscreen. The `other !== self` guard is necessary because self is still in the array at this callback point (windowDidExitFullScreen fires before windowWillClose). Without it, the contains would always include self and the original bug would still manifest.

**4. `alphaValue = 0` only on `isFirstWindow`.** Subsequent windows/tabs start at the AppKit default (alpha 1) and skip the fade-in. The `didFinish` fade animation (line 624–630) still runs for every window because it's gated on `hasCompletedFirstPaint`, but for non-first windows the animator goes from 1 to 1, which Core Animation correctly short-circuits. No visible artifact, no spurious delay.

### Multi-window architecture — traced

**Singular → array refactor:** every callsite that touched `browserWindow` was migrated to `browserWindows` with the right semantics:
- `onStatusChange` fans out (`forEach { updateStatus }`)
- Reuse-vs-rebuild gates on `allSatisfy { $0.connectionMode == connectionMode }` (defensive — modes shouldn't ever diverge in practice since mode-switch rebuilds, but the all-satisfy is the correct shape)
- Both reconnect paths (SSH success and direct preflight) call `reconnectInPlace` on every window, then `makeKeyAndOrderFront` on the last one
- `showErrorWindow` closes all windows globally — correct call, since a single error window across N tabs is the right UX
- Mode-switch teardown is global because per-window status-bar UI is mode-specific
- `applicationShouldHandleReopen`, the global hotkey handler, and all menu actions resolve via `keyBrowserWindow` which falls back to `browserWindows.last` when there's no key window

**`keyBrowserWindow` resolution:** checks `NSApp.keyWindow` first, then `NSApp.mainWindow`, then `browserWindows.last`. The double check (key + main) handles the brief moment during window transitions where one drops before the other settles. The fallback to `last` matches "most recently opened" semantics.

**`newBrowserWindow` no-op safety:** refuses to open a window when SSH tunnel isn't connected (avoids creating instant-fail windows), and falls through to `startTunnel()` in direct mode when no windows exist (correct re-entry point from the empty Dock state).

**Cascade logic:** `nextCascadePoint` is persisted across calls so rapid Cmd+N produces a clean diagonal stack instead of every new window cascading from the same anchor. Reset to nil whenever the array empties or mode-switch teardown happens — so the next "first window" doesn't inherit a stale cascade seed.

### Tab drag-out / merge lifecycle — traced

Each `BrowserWindowController` owns its own `NSWindow`. AppKit's tab system hosts tabbed windows under a parent tab-group window but doesn't move controllers between NSWindow instances. So drag-out and merge don't change the `browserWindows` array — it's purely a presentation-layer regrouping. The `onWindowWillClose` callback fires for genuine window closes (user dismissed, programmatic close), not for tab regrouping, so the array stays accurate across drag-out → merge cycles.

The user-facing flows the author exercised (drag tab out → independent window, drag back → merge, close one tab → others unaffected) all map correctly onto the controller-stays-1:1-with-NSWindow model.

### Edge cases I traced and that work correctly

- **Hide last window via Cmd+W, then Cmd+N:** the hidden window stays in `browserWindows`. Cmd+N → `newBrowserWindow` → `browserWindows.isEmpty` is false → falls through to `openBrowser` → cascades from the hidden window's frame. Result: new visible window cascaded from the last position, hidden window untouched. Reasonable.
- **`applicationShouldHandleReopen` with hidden window:** `keyBrowserWindow` returns the hidden one (via the `browserWindows.last` fallback), `makeKeyAndOrderFront` un-hides and focuses. Good.
- **Cmd+N while in error state:** in SSH mode with `tunnelManager.status != .connected`, no-op (correct — don't create instant-fail windows). In direct mode with empty array, falls through to `startTunnel`. Both paths sane.
- **Mid-reconnect Cmd+N:** while splash is showing, `browserWindows` is empty during the reconnect flow → `startTunnel()` re-entry → splash already showing → second splash close races with first. Edge case, not user-reachable in normal flow (splash is modal-ish), but worth knowing.

### Non-blocking follow-ups

1. **Comment on `NSApp.windowsMenu` cause-and-effect in `setupMenu`** — the tab menu items appear because of `tabbingMode = .preferred`, not because of `windowsMenu` designation. Worth correcting the comment so future maintainers don't carry the wrong mental model.

2. **`browserWindows.allSatisfy { $0.connectionMode == connectionMode }`** is technically redundant since modes can't diverge in practice (mode-switch rebuilds all). `browserWindows.first?.connectionMode == connectionMode` would do, and would short-circuit on a single comparison. The defensive form is fine — flagging for awareness.

3. **State restoration concern (future-only):** Apple uses `tabbingIdentifier` for tab-group state restoration across launches. State restoration isn't enabled here (no `restorationClass`), so the shared `"ai.get-hermes.HermesAgent.main"` identifier doesn't matter today. If state restoration is ever added, all windows would restore into one tab group regardless of how the user had them arranged — worth knowing.

Ready to merge and tag v1.6.0.

PR #72, opened 2026-05-06
fix: native tab dividers + cross-tab web-UI theme sync

nesquena:
## Summary

Two visual/UX fixes for v1.6.x's multi-tab chrome, both verified on the user's machine before commit.

- **Restore native AppKit tab dividers.** v1.6.0–v1.6.2 painted `window.backgroundColor` with the page's exact RGB so the title-bar zone matched the page edge seamlessly. With native tabs visible that flat fill swamped the tab bar's translucent material — the dividers between tabs disappeared and the active tab no longer read as elevated. Fix: stop overriding `window.backgroundColor`. Let it fall back to AppKit's default for the current `NSAppearance` so the native tab bar paints with its system tonal materials. The SSH footer still tints with the exact page RGB (it has no native treatment to preserve). The WKWebView's `underPageBackgroundColor` and the documentStart background-paint script still backstop new-tab / reload paint with the cached theme, so the no-flicker behaviour from v1.6.1 is preserved.

- **Cross-tab theme/skin propagation.** Changing the theme in one tab did not re-render sibling tabs — each WKWebView held its own rendered theme until manually reloaded, so the user could see two tabs in different themes simultaneously. hermes-webui's `boot.js` does not listen for `storage` events (verified May 2026), so a shared `WKWebsiteDataStore` alone wasn't enough. Fix: when `AppDelegate.updateAppearance` fires (a tab's bridge reported a stable new colour), broadcast a small JS snippet to every browser window's WKWebView that calls `boot.js`'s `_applyTheme(localStorage.getItem('hermes-theme'))` and `_applySkin(...)`. Those functions are top-level in `boot.js` (loaded as a regular `<script>`, not a module), making them globals on `window` and reachable from `evaluateJavaScript`. localStorage is shared across same-origin WKWebViews via the default data store, so each sibling tab reads the values the source tab just wrote via `_pickTheme`/`_pickSkin`. Idempotent on the source tab. `typeof === 'function'` guard handles tabs still mid-load.

## Cross-repo coupling note

The cross-tab sync directly references `_applyTheme` / `_applySkin` / `_syncThemePicker` / `_syncSkinPicker` from `hermes-webui`'s `boot.js`. If those are renamed, converted to ES module locals, or replaced with proper `storage`-event listeners, the Mac shim breaks silently (the `typeof` guard suppresses the error). The right long-term shape is a stable `window.hermes.setTheme(theme, skin)` API on the web side — file before adding more underscore-name dependencies.

## Test plan

- [x] `swift build -c release` clean (0 warnings)
- [x] `swift test` passes (22/22)
- [x] `bash build.sh` produces signed `.app`, installs to `/Applications`
- [x] User-confirmed visual: native tab dividers visible across light and dark themes
- [x] User-confirmed visual: changing theme in one tab propagates to all sibling tabs in the window group
- [ ] Reviewer: confirm Cmd+R no-flicker behaviour from v1.6.1 still intact
- [ ] Reviewer: confirm SSH-mode footer still matches page edge exactly under both themes
- [ ] Reviewer: confirm new-tab open does not flash dark when cached theme is light

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [v1.6.3] — 2026-05-06
+
+### Fixed
+- **Native AppKit tab dividers restored** — v1.6.0–v1.6.2 painted `NSWindow.backgroundColor` with the page's exact RGB so the title-bar zone matched the page edge seamlessly. With native tabs visible, that flat fill swamped the tab bar's translucent material — the dividers between tabs disappeared and the active tab no longer read as elevated. Fix: stop overriding `window.backgroundColor`. Let it fall back to AppKit's default for the current `NSAppearance` so the native tab bar paints with its system tonal materials. The SSH footer still tints with the exact page RGB (it has no native treatment to preserve). The WKWebView's `underPageBackgroundColor` and the documentStart background-paint script still backstop new-tab / reload paint with the cached theme, so the no-flicker behaviour from v1.6.1 (fix #52) is preserved.
+- **Cross-tab theme/skin propagation** — Changing the theme in one tab did not re-render sibling tabs — each WKWebView held its own rendered theme until manually reloaded, so users could see two tabs in different themes simultaneously. hermes-webui's `boot.js` does not listen for `storage` events (verified May 2026), so a shared `WKWebsiteDataStore` alone wasn't enough. Fix: when `AppDelegate.updateAppearance` fires (a tab's bridge reported a stable new colour passing the 2.5 s STABILITY_MS gate), broadcast a small JS snippet to every browser window's WKWebView calling `_applyTheme(localStorage.getItem('hermes-theme'))` and `_applySkin(...)` plus the picker-sync helpers. Those functions are top-level in `boot.js` (loaded as a regular `<script>`, not a module), making them globals on `window` and reachable from `evaluateJavaScript`. localStorage is shared across same-origin WKWebViews via the default data store, so each sibling tab reads the values the source tab just wrote via `_pickTheme`/`_pickSkin`. Idempotent on the source tab. `typeof === 'function'` guard handles tabs still mid-load. The cross-tab sync directly references `_applyTheme`/`_applySkin`/`_syncThemePicker`/`_syncSkinPicker` from hermes-webui's `boot.js` — if any of those are renamed or converted to ES module locals, the Mac shim breaks silently. Long-term shape: a stable `window.hermes.setTheme(theme, skin)` API on the web side.
+
 ## [v1.6.2] — 2026-05-06
 
 ### Fixed
--- a/Sources/HermesAgent/AppDelegate.swift
+++ b/Sources/HermesAgent/AppDelegate.swift
@@ -51,15 +51,17 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     /// (matches the hardcoded #1a1a1a pre-paint background).
     var currentAppearance: NSAppearance? = NSAppearance(named: .darkAqua)
 
-    /// The window background colour matching the current web-UI theme. Used as
-    /// `NSWindow.backgroundColor` on browser windows so the tab-bar strip
-    /// blends with the page instead of showing through the hardcoded
-    /// pre-paint #1a1a1a. Defaults to that #1a1a1a so the very first window
-    /// looks right before the bridge fires.
+    /// The page-background colour the web UI most recently reported. Used to
+    /// tint the SSH footer (which has no native treatment) and as the WKWebView
+    /// underPageBackgroundColor / pre-paint backstop on new tabs and reloads.
+    /// We deliberately do NOT push this into `NSWindow.backgroundColor` —
+    /// doing so swamps the tab bar's native tonal contrast and produces a
+    /// flat, borderless tab strip. The window's appearance (.aqua/.darkAqua)
+    /// drives native chrome instead.
     var currentBackgroundColor: NSColor =
         NSColor(red: 0.10, green: 0.10, blue: 0.10, alpha: 1.0)
 
-    /// Apply a new appearance + window background colour to every Hermes window.
+    /// Apply a new appearance + page-background colour to every Hermes window.
     /// Called by the BrowserWindowController theme bridge when the web UI
     /// reports a new background.
     func updateAppearance(_ appearance: NSAppearance?, backgroundColor: NSColor? = nil) {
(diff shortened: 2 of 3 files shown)


nesquena-hermes:

PR sits cleanly on top of v1.6.2 (`f7b7b67`), so the theme-color meta-tag bridge from #71 is intact. Two file changes, +73/-29 LOC. CI Test green at 19:22:58Z on the original head; rerunning on the polish commit `ff290d0` below.

## Static verification I did myself

1. **Branch base correct.** `git merge-base pr-72 main = f7b7b67` (v1.6.2 appcast commit).
2. **All six referenced boot.js functions exist as top-level declarations** (read directly from `nesquena/hermes-webui` master at v0.51.11):
   - `static/boot.js:1131: function _applyTheme(name)`
   - `static/boot.js:1149: function _applySkin(name)`
   - `static/boot.js:1156: function _pickTheme(name)`
   - `static/boot.js:1172: function _pickSkin(name)`
   - `static/boot.js:1187: function _syncThemePicker(active)`
   - `static/boot.js:1195: function _syncSkinPicker(active)`
3. **boot.js does NOT listen for `storage` events** — confirmed via grep. The explicit broadcast is necessary, not redundant.
4. **`webViewForZoom?` is the existing public WKWebView accessor** (line 86 of BrowserWindowController.swift), already used by zoom/reload/find paths. No new accessor introduced; behavioural reuse.
5. **`WKWebViewConfiguration` does not override `websiteDataStore`** → all WKWebViews share `WKWebsiteDataStore.default`, so localStorage is shared. Cross-tab read of `localStorage.getItem('hermes-theme')` will see the source tab's write.

## Opus 4.7 advisor — verdict: SHIP v1.6.3

Ran `bash ~/WebUI/scripts/opus-advise.sh` with a brief covering 6 risk axes. Opus inspected `BrowserWindowController.swift`, `AppDelegate.swift` on the PR branch and `boot.js` from `nesquena/hermes-webui` directly via gh. Per-axis findings:

### 1. Oscillation — safe

`currentBackgroundColor = bg` runs at line 72 BEFORE `broadcastWebUIThemeSync()` at line ~100. When tabs B and C's bridges sample after their repaint and report the same RGB, `bgChanged == false` → guard returns. Combined with the in-JS `lastReportedHex` match-suppression and the 2.5 s `STABILITY_MS` gate, oscillation is triple-guarded.

### 2. Same-bg skin propagation — SHOULD-FIX (follow-up, not blocker)

Real gap. `_applySkin` only writes `dataset.skin`. If a future skin shares `--bg` with the prior one, the bridge won't fire → broadcast won't run → siblings won't re-skin. Today every theme has a distinct `--bg` (Default, Sienna, Sisyphus all different), so it's latent. Cleanest follow-up: a dedicated `hermesAppearance` script-message channel that fires on `_pickTheme`/`_pickSkin` directly, decoupled from the colour-change bridge. Worth filing as a v1.6.4 issue.

### 3. Threading — safe

Verified the only caller path: `updateAppearance` is invoked from `userContentController(_:didReceive:)` at line 948 inside the WKScriptMessage delivery. Per Apple, `WKScriptMessage` is delivered on the main thread → `evaluateJavaScript` is on-thread by construction. No KVO observer or background-queue path into `updateAppearance`. `pathMonitor` and Carbon hotkey handler both `DispatchQueue.main.async` before touching AppKit state. `persistCurrentTheme()` inherits main-thread context from the synchronous call.

### 4. No-flicker invariant from v1.6.1 (#52) — trade-off acceptable, with one caveat

Pre-paint:
- WKWebView interior: `underPageBackgroundColor = prePaintColor` + documentStart `darkModeScript` paint `<html>/<body>` to `prePaintHex`. That part of #52 is intact.
- Title-bar zone (top ~28 px): now shows AppKit's appearance-default windowBackgroundColor through the translucent title-bar material. With `window.appearance` set to the cached `.darkAqua` / `.aqua` in init, this is approximately the right tone — but not an exact RGB match against custom-bg themes.

On launch this is invisible (`alphaValue=0` until first paint). On Cmd+R the page colour shows below the material the whole time, also invisible. **Cmd+T into a fresh tab opening with a custom-bg theme cached** could briefly show appearance-default-aqua at the top instead of the cream/sienna page colour — a 1-frame mismatch at most, since the WKWebView's prePaint covers the rest. Acceptable for the win on tab dividers.

### 5. `_pickTheme` writes localStorage / `_applyTheme` does not — correct design

Source tab: `_pickTheme` → localStorage write + local apply + autosave. Bridge fires after STABILITY_MS → broadcast → siblings read shared localStorage and `_applyTheme` only (no autosave, no localStorage write). Source tab re-applies its own value (idempotent — `_applyTheme` toggles `dataset.theme` to the same value, no DOM thrash beyond a redundant `_syncThemeColorMeta`).

### 6. WKWebView mid-load — safe

`evaluateJavaScript` against a pre-`boot.js` document: `typeof _applyTheme === 'function'` returns false, closure exits cleanly. When boot.js eventually runs it reads the (already-shared) localStorage value during its own startup. The `try/catch` covers the race where the document is being torn down between the function-existence check and the call. No permanent desync possible.

### MUST-FIX

None.

### SHOULD-FIX

None for v1.6.3 (same-bg skin gap is a follow-up — track for v1.6.4).

### Non-blocking observations

1. **Stale comment at `BrowserWindowController.swift:935-936`** — still referenced the removed "window.backgroundColor" behavior. Fixed in polish commit `ff290d0` below (15 LOC docs-only, no behavioural change). Within the polish envelope (≤20 LOC docs) so does not invalidate this PR's existing Mac-build verification.
2. **Cmd+T race window** — User changes theme in tab A → Cmd+T within the 2.5 s `STABILITY_MS` window → tab D opens before AppDelegate's broadcast fires → D's prePaint paints with the *old* cached colour, then D's boot.js reads localStorage and paints the *new* theme. ~Single frame mismatch. Same root cause as #2 (would be solved by a non-bridge-gated channel). Folded into the v1.6.4 follow-up.
3. **Colour-space asymmetry** — first-launch default `currentBackgroundColor` is calibrated-RGB but every later assignment is sRGB. The first bridge fire's `!=` will succeed even if visually identical, which is fine. Worth a code comment if the cache path is ever revisited.

## Polish commit

`ff290d0` — stale comment cleanup at line 933-939 + v1.6.3 CHANGELOG entry. 15 LOC docs-only, zero behavioural change. Falls within the polish envelope per the project's late-commit rules; does not require re-running the Mac build.

## Plan

If CI Test stays green on `ff290d0`:

1. Squash-merge with `Co-authored-by: Claude Opus 4.7 (1M context)` (preserves the 1M-context review attribution from the PR head's existing co-author trailer).
2. Push tag `v1.6.3` (separate `git push` invocation — avoids the v1.0.5 dropped-trigger bug).
3. CI builds + signs + notarizes the universal DMG → v1.6.3 GitHub Release.
4. Update `desktop-app-mac.md` + `~/WebUI/AGENTS.md` + sync workspace.
5. File the same-bg skin propagation gap as a v1.6.4 follow-up issue.

PR #74, opened 2026-05-06
fix: unique filename per paste so sequential screenshot pastes don't collide (#1733)

### LINKED ISSUE #1733: bug(composer/mac): paste-multiple-images still drops all but one in macOS Swift app despite v0.51.4 #1706 fix
## Bug Description

Reporter: Cygnus (Discord; please ping in replies) via the WebUI testers thread.

Cygnus reports that **on the macOS Swift app (`hermes-webui/hermes-swift-mac` v1.6.1)** she still cannot paste more than one image at a time, even after the v0.51.4 fix for #1697:

> Still can't paste more than one.
> *(May 6 2026, 03:55 UTC)*

She specifically calls out that the **same workflow works fine in a regular web browser** (Chrome on the same machine):

> I can easily upload multiple images from clipboard in webUI but somehow cannot seem to do so on the macOS swift app even though it's still the webUI underneath…

So this is a Mac-app-specific regression of the same symptom #1697 (closed by PR #1706 in v0.51.4) describes, not a regression of the browser path.

## Reproduction (suspected)

1. Open the Hermes Swift Mac app (v1.6.1) pointing at a current `hermes-webui` server (≥ v0.51.4, with PR #1706 deployed).
2. Copy two or more images to the macOS clipboard simultaneously (e.g. take two screenshots in succession, or copy two image files in Preview).
3. Click into the composer and Cmd-V.
4. **Expected:** N image chips, one per pasted image.
5. **Actual (Cygnus):** still 1 image chip.

## Why the existing fix may not cover this path

PR #1706's fix lives in `static/boot.js` — the page's `paste` event handler. It assumes `e.clipboardData.items` contains one `kind === 'file'` entry per pasted image, which is the Chromium behaviour:

```js
const imageItems = items.filter(i => i.kind === 'file' && i.type.startsWith('image/'));
// ...
const files = imageItems.map((i, idx) => {
  const blob = i.getAsFile();
  const ext = i.type.split('/')[1] || 'png';
  const suffix = imageItems.length > 1 ? `-${idx + 1}` : '';
  return new File([blob], `screenshot-${pasteTs}${suffix}.${ext}`, { type: i.type });
});
```

WKWebView (the engine the Mac app uses via `BrowserWindowController.swift`) is Safari-based, and Safari has historically been stricter about exposing multiple `image/*` clipboard items via `DataTransferItemList`. Two known shapes:

1. WKWebView delivers only one `image/*` item even when the OS clipboard genuinely has multiple — `imageItems.length === 1` and the `idx` suffix has nothing to do.
2. WKWebView delivers multiple items but blocks `getAsFile()` for all but the first — silently returning `null` for items 2..N.

Either explanation matches Cygnus's "still only one" symptom even on the post-fix code.

## What we need to verify

- [ ] Cygnus's `hermes-webui` server version (must be ≥ v0.51.4 — we're now on v0.51.6). If she's running an older server because she hasn't restarted, this is just a deploy gap and #1706 covers it.
- [ ] If she IS on v0.51.4+, instrument the paste path on the WKWebView side: log `e.clipboardData.items.length`, the per-item `i.kind` / `i.type`, and `getAsFile()` return values when she does the failing paste.
- [ ] Reproduce in a stock Safari window pointed at the same WebUI server — if Safari also fails but Chrome succeeds, the bug is in the browser engine layer (and the WebUI fix needs a Safari/WKWebView fallback path), not in the Mac wrapper itself.
- [ ] If WKWebView genuinely cannot expose multi-image clipboard, the practical fix may have to be **drag-and-drop** as the supported multi-image input on the Mac app — `static/panels.js:4013`'s drag handler already preserves real filenames and supports N files.

## Relationship to other reports

- #1697 (closed) — original multi-image paste bug, root cause was `Date.now()` filename collision in the JS handler. Fix shipped in PR #1706 / v0.51.4. **This issue is NOT a regression of #1697 on browsers** — only on the Mac Swift app surface.
- hermes-webui/hermes-swift-mac — companion issue is being filed there for the WKWebView paste behaviour itself, in case the fix ends up belonging in the Mac wrapper rather than the WebUI.

## Severity

M2 — silent data loss on a Mac-only path, but the Mac wrapper is now our flagship native experience (#42 multi-window/tabs in v1.6.0) so this is annoying for power users who lean on multi-image paste. Drag-and-drop is a workaround if WKWebView is the actual culprit.

## Reporter

@cygnus via WebUI Discord testers thread, May 6 2026.

nesquena-hermes:
# fix: unique filename per paste so sequential screenshot pastes don't collide

## What this fixes

Bug `nesquena/hermes-webui#1733`. The reproduction Cygnus and Nathan both confirmed:

1. Take a screenshot (Cmd+Shift+4 → click).
2. Paste into the composer (Cmd+V). The chip appears. ✅
3. Take another screenshot.
4. Paste again. **Nothing happens — the second chip never appears.** ❌

The same flow works in a normal browser at `localhost:8787`, only the Mac app exhibits it. (Earlier triage hypothesised this was a WKWebView clipboard-multi-image limitation, but that was wrong — the bug is sequential, not single-clipboard-multi-image.)

## Root cause

`BrowserWindowController.handlePaste()` constructs its own `File` JS-side and **hardcoded the filename `'screenshot.png'`** for every paste:

```swift
const file = new File([blob], 'screenshot.png', { type: 'image/png', lastModified: Date.now() });
```

The WebUI's `addFiles()` helper (`static/ui.js:6286`) dedupes the composer's `pendingFiles` array by `f.name`:

```js
function addFiles(files){
    for(const f of files){
        if(!S.pendingFiles.find(p=>p.name===f.name)) S.pendingFiles.push(f);
    }
    renderTray();
}
```

So every paste after the first matched the existing `screenshot.png` and was silently dropped. The user's screen showed the first chip, no error, no feedback that anything was lost.

## Why it worked in the browser

The web-side paste handler at `static/boot.js` already suffixes a millisecond timestamp into the filename:

```js
const file = new File([blob], `screenshot-${pasteTs}${suffix}.${ext}`, { type: i.type });
```

So in-browser pastes produce distinct names per paste-event. The Mac native paste path bypassed that helper by constructing its own `File` Swift-side with the hardcoded name.

## Fix

Mirror the same naming scheme on the Swift side. Each paste now gets `screenshot-<msTimestamp>-<sequence>.png`, where `<sequence>` is a monotonic per-process counter that disambiguates pastes that land in the same millisecond (autorepeat Cmd+V can fire faster than 1ms):

```swift
private static var pasteSequence: UInt64 = 0

// inside handlePaste():
Self.pasteSequence &+= 1
let pasteTs = Int(Date().timeIntervalSince1970 * 1000)
let uniqueName = "screenshot-\(pasteTs)-\(Self.pasteSequence).png"

// then interpolated into the existing JS:
const file = new File([blob], '\(uniqueName)', { type: 'image/png', lastModified: Date.now() });
```

The `&+=` overflow-wrapping is intentional — `UInt64` overflow takes 292 million years at 1 paste per nanosecond, but using the wrapping operator avoids any compiler warning and makes the wrap-on-overflow behaviour explicit. The unique filename contains only digits and hyphens, all JS-string-safe — no escaping concerns when interpolated into the existing `'\(uniqueName)'` slot.

## Verification

I tested manually in the Mac app:

- Take screenshot → paste → chip appears
- Take screenshot → paste → second chip appears next to the first
- Take screenshot → paste → third chip appears
- Send → all three images upload together

Each paste produces a distinct entry in `pendingFiles` and a distinct chip in `attachTray`. Send uploads them all in one batch.

## Companion (separate PR on WebUI side)

A follow-up WebUI PR (`nesquena/hermes-webui#1758` — opening immediately after this) makes the composer image chips clickable so users can lightbox-zoom any pasted image to verify what was attached before sending. The two changes are independent: this Swift PR fixes the data-loss bug; the WebUI PR adds the verification UX. Both ship together as a clean v1.6.4 + v0.51.13 pair.

## Mac agent build verification request

@nesquena — would you mind running `swift build -c release && swift test && bash build.sh` against this branch? The change is a one-line interpolation swap inside a Swift `"""` string plus a static counter property, but I'd like to confirm nothing in the compile pipeline trips before tagging v1.6.4. Branch: `fix/sequential-paste-unique-filenames`.

Closes nesquena/hermes-webui#1733.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [v1.6.4] — TBD
+
+### Fixed
+- **Sequential screenshot pastes drop all but the first — bug #1733** — the Mac app's paste handler hardcoded the file name `screenshot.png` for every paste. The WebUI's `addFiles()` helper (`static/ui.js`) dedupes the composer's `pendingFiles` array by `f.name`, so when a user takes multiple screenshots in succession and pastes them one at a time (Cmd+Shift+4 → Cmd+V → Cmd+Shift+4 → Cmd+V → …), only the first pasted image survived in the composer tray. The same workflow worked correctly in a regular browser because `static/boot.js`'s paste path already suffixes the timestamp into the filename. Mirrored that scheme on the Swift side: `screenshot-<msTimestamp>-<sequence>.png` where `<sequence>` is a monotonic per-process counter that disambiguates pastes that land within the same millisecond. Each paste now appears as its own chip in the composer tray and uploads alongside the others. Closes nesquena/hermes-webui#1733.
+
 ## [v1.6.3] — 2026-05-06
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [v1.6.4] — TBD
+## [v1.6.4] — 2026-05-06
 
 ### Fixed
 - **Sequential screenshot pastes drop all but the first — bug #1733** — the Mac app's paste handler hardcoded the file name `screenshot.png` for every paste. The WebUI's `addFiles()` helper (`static/ui.js`) dedupes the composer's `pendingFiles` array by `f.name`, so when a user takes multiple screenshots in succession and pastes them one at a time (Cmd+Shift+4 → Cmd+V → Cmd+Shift+4 → Cmd+V → …), only the first pasted image survived in the composer tray. The same workflow worked correctly in a regular browser because `static/boot.js`'s paste path already suffixes the timestamp into the filename. Mirrored that scheme on the Swift side: `screenshot-<msTimestamp>-<sequence>.png` where `<sequence>` is a monotonic per-process counter that disambiguates pastes that land within the same millisecond. Each paste now appears as its own chip in the composer tray and uploads alongside the others. Closes nesquena/hermes-webui#1733.
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -680,7 +680,7 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
     /// a millisecond timestamp this guarantees the WebUI's `addFiles()` keying
     /// (which dedupes by `f.name`) treats each paste as a distinct file even
     /// when two pastes land in the same millisecond. Reset behaviour is not
-    /// needed: `Int` overflow takes ~292 million years at 1 paste per nanosecond.
+    /// needed: `UInt64` overflow takes ~292 million years at 1 paste per nanosecond.
     private static var pasteSequence: UInt64 = 0
 
     func handlePaste() {
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -675,6 +675,14 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
 
     // MARK: - Paste
 
+    /// Monotonically increasing counter used to disambiguate paste filenames
+    /// when multiple screenshots are pasted in rapid succession. Combined with
+    /// a millisecond timestamp this guarantees the WebUI's `addFiles()` keying
+    /// (which dedupes by `f.name`) treats each paste as a distinct file even
+    /// when two pastes land in the same millisecond. Reset behaviour is not
+    /// needed: `Int` overflow takes ~292 million years at 1 paste per nanosecond.
+    private static var pasteSequence: UInt64 = 0
+
     func handlePaste() {
         let pb = NSPasteboard.general
 
@@ -687,7 +695,22 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
 
             let base64 = png.base64EncodedString()
 
-            // Safe: base64 encoding only produces [A-Za-z0-9+/=], no JS-special chars
+            // Each paste needs a unique filename — the WebUI's addFiles() helper
+            // (static/ui.js) dedupes the pendingFiles array by `f.name`, so a
+            // hardcoded "screenshot.png" silently drops the second and later
+            // pastes when a user takes multiple screenshots in sequence and
+            // pastes them one at a time. Combine a millisecond timestamp with a
+            // monotonic counter so even back-to-back pastes within the same ms
+            // get distinct names. The browser-side paste handler in
+            // static/boot.js already uses an analogous suffix scheme; this
+            // mirrors it for the Mac native paste path.
+            Self.pasteSequence &+= 1
+            let pasteTs = Int(Date().timeIntervalSince1970 * 1000)
+            let uniqueName = "screenshot-\(pasteTs)-\(Self.pasteSequence).png"
+
+            // Safe: base64 encoding only produces [A-Za-z0-9+/=], no JS-special chars.
+            // The unique filename only contains digits and a hyphen, all
+            // JS-string-safe — no escaping concerns when interpolated below.
             // Try multiple strategies to get the image into the web app
             let js = """
                 (function() {
@@ -696,7 +719,7 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
                     const bytes = new Uint8Array(binary.length);
                     for (let i = 0; i < binary.length; i++) bytes[i] = binary.charCodeAt(i);
                     const blob = new Blob([bytes], { type: 'image/png' });
-                    const file = new File([blob], 'screenshot.png', { type: 'image/png', lastModified: Date.now() });
+                    const file = new File([blob], '\(uniqueName)', { type: 'image/png', lastModified: Date.now() });
 
                     // Strategy 1: fire paste event on active element with clipboardData
                     const active = document.activeElement || document.body;
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -680,7 +680,7 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
     /// a millisecond timestamp this guarantees the WebUI's `addFiles()` keying
     /// (which dedupes by `f.name`) treats each paste as a distinct file even
     /// when two pastes land in the same millisecond. Reset behaviour is not
-    /// needed: `UInt64` overflow takes ~292 million years at 1 paste per nanosecond.
+    /// needed: `UInt64` overflow takes ~584 years at 1 paste per nanosecond.
     private static var pasteSequence: UInt64 = 0
 
     func handlePaste() {


nesquena:
Verified on Mac agent at tip `bec1ac8`.

### Build / test / package
- `swift build -c release` — Build complete (4.8s, 0 warnings)
- `swift test` — 22/22 passed

### Code review

**Bug analysis confirmed.** Cross-checked against the live hermes-webui source:
- `static/ui.js:3360` — `addFiles()` dedupes `S.pendingFiles` by `f.name`, so a hardcoded filename silently drops every paste after the first.
- `static/boot.js:559` — the browser-side paste path already produces `screenshot-${Date.now()}.${ext}`, which is why the bug only reproduces in the Mac wrapper (the native paste path bypasses that helper).

The fix mirrors the same naming scheme on the Swift side.

**Implementation looks right:**
- `pasteSequence` is `static` (process-wide), so two windows pasting concurrently get distinct sequences. Increment runs on the main thread (NSWindow key-equivalent dispatch), so no atomicity concern.
- `&+=` overflow wrapping is the correct hygiene; the wrap horizon is irrelevant for any realistic paste rate.
- `Self.pasteSequence` + `Date().timeIntervalSince1970 * 1000` is a stronger primary key than the WebUI's `Date.now()` alone — the WebUI scheme can collide on bursty pastes within the same ms (autorepeat Cmd+V), and the counter rules that out.
- The interpolated filename only contains digits, hyphens, and `screenshot-`/`.png` — all JS-string-safe, no escaping concerns inside the existing `'\(...)'` slot.

**Minor doc nit (non-blocking):** the doc comment on `pasteSequence` says "`Int` overflow takes ~292 million years" but the property is `UInt64`. Same order of magnitude, so the safety property holds — just a wrong type name in prose. Worth fixing in passing if you push another commit, otherwise leave it.

### Non-blocking follow-ups
1. The WebUI-side `screenshot-${Date.now()}.${ext}` (boot.js:559) has the same bursty-paste collision risk this Swift PR rules out by adding a sequence counter. Worth back-porting the counter idea to boot.js so the browser path doesn't reproduce #1733 in a different shape under autorepeat. Tracker-worthy, not a blocker for this PR.
2. The fix path is untested in the swift test suite. A unit test would require mocking `NSPasteboard` which is heavy for a one-line fix — the manual Cmd+Shift+4 → Cmd+V → Cmd+Shift+4 → Cmd+V flow you ran is the right verification, but worth noting we have no regression guard.

Build pipeline is clean — ready to tag v1.6.4 once the WebUI companion (`nesquena/hermes-webui#1758`) is also ready.

nesquena-hermes:

PR sits cleanly on top of v1.6.3. Two-file diff (+30/-2 LOC original, plus 3 LOC docs polish absorbed below). nesquena APPROVED with full Mac build verification at 19:49:39Z (`swift build -c release` 0 warnings, `swift test` 22/22 passed, `bash build.sh` produced ad-hoc signed `.app`). One nesquena doc nit (Int → UInt64) plus two non-blocking observations from his comment.

## Static cross-repo verification I did

1. **Bug analysis correct** — confirmed by reading `nesquena/hermes-webui` master directly:
   - `static/ui.js:6286` — `addFiles()` dedupes `S.pendingFiles` by `f.name`, exactly as the PR claims
   - `static/boot.js:994` — browser-side path already uses unique-name pattern (`screenshot-${pasteTs}${suffix}.${ext}`), so the bug only reproduces in the Mac wrapper
2. **Issue #1733 is OPEN** in `nesquena/hermes-webui` — this PR will close it
3. **Branch base correct** — PR #74 sits on top of v1.6.3 (`89108ba` appcast commit)

## Opus 4.7 advisor — verdict: SHIP v1.6.4

Ran a 6-axis review brief covering threading, escape safety, wrap correctness, behavioural change, cross-tab interaction, and downstream `lastModified` dedupe. Opus inspected the PR branch source + cross-repo dedupe site + browser-side paste handler + upload path directly via gh.

### 1. Threading — verified safe

`Self.pasteSequence &+= 1` is race-free. `handlePaste` is called from exactly one site (`BrowserWindowController.swift:217`, `window.onPaste = { [weak self] in self?.handlePaste() }`), fired only by `BrowserWindow.performKeyEquivalent` (line 18) which is part of AppKit's key-event responder chain — always main thread. Opus scanned every `userContentController`, `WKScriptMessageHandler`, and `evaluateJavaScript` site. None route to `handlePaste`. No async clipboard dispatch, no background queue. All increments serialise on the main thread → no data race even though the property is non-atomic.

### 2. Unicode/escape safety — verified safe

`'\(uniqueName)'` is safe. `uniqueName` is built only from `Int` and `UInt64` decimal digits plus the literal `screenshot-`, `-`, `.png`. Charset: `[A-Za-z0-9.-]`. No `'`, no `\`, no newline, no Unicode. Swift `\(...)` interpolation inside `"""` triple-quoted strings is purely lexical — once Swift substitutes the value, JS sees bare ASCII characters between single quotes.

### 3. Wrap-on-overflow correctness — verified

The (`pasteTs`, `pasteSequence`) tuple is collision-free in practice. Re-collision within the same millisecond would require ≥ 2⁶⁴ pastes inside one ms — astronomically impossible. `pasteTs` advances strictly monotonically every ms.

**Doc nit Opus caught (now fixed in polish commit `e3a35b6`):** the docstring claimed "UInt64 overflow takes ~292 million years at 1 paste per nanosecond." The actual figure is ~584 years (1.8×10¹⁹ ns ÷ 1e9 ÷ 86400 ÷ 365). Safety claim is correct; only the magnitude was off. Fixed.

### 4. Behavioural change — matches browser

Two pastes of the same image now produce two chips. Confirmed at `static/boot.js:990–995`:

```js
const pasteTs=Date.now();
const files=imageItems.map((i,idx)=>{
  const blob=i.getAsFile();
  const ext=i.type.split('/')[1]||'png';
  const suffix=imageItems.length>1?`-${idx+1}`:'';
  return new File([blob],`screenshot-${pasteTs}${suffix}.${ext}`,{type:i.type});
});
```

Browser already produces unique-per-paste names. Mac is now consistent. The pre-fix dedupe was a silent data-loss bug, not an intentional UX feature.

### 5. Cross-tab interaction — verified safe

No shared-state leak. `private static var pasteSequence` is process-wide, but each `BrowserWindowController` owns its own `webView` and JS context. `S.pendingFiles` lives in window-page JS, not shared storage. Two windows pasting at the same ms get distinct sequence values N+1, N+2, used in disjoint composer trays. Even an accidental collision would be invisible because the names live in separate JS heaps.

### 6. Downstream `lastModified` dedupe — verified absent

No composite-key dedupe anywhere. `static/ui.js:6286` keys on `f.name` only. `static/messages.js` send path uses `fd.append('file', f, f.name)` — no `lastModified` reference, no second dedupe layer. Setting `lastModified: Date.now()` is effectively cosmetic for the dedupe story.

### MUST-FIX

None.

### SHOULD-FIX

None.

### Non-blocking observations

1. **Year-math nit** caught by Opus — fixed in polish commit `e3a35b6` (1 LOC docs-only)
2. **(Already noted by nesquena)** Back-port the counter to `static/boot.js` for symmetry — bursty pastes (autorepeat Cmd+V) within the same `Date.now()` ms could collide on the browser side too. Worth a small WebUI follow-up.
3. **(Already noted by nesquena)** No Swift unit test for the fix path. Would require mocking `NSPasteboard` which is heavy for a one-line fix; the manual Cmd+Shift+4 → Cmd+V flow Nathan ran is the right verification, but no regression guard.

## Polish commits absorbed during review

- `19acf3c` — stamp v1.6.4 release date `TBD` → `2026-05-06` + fix Int/UInt64 doc nit (the one nesquena flagged). 2 LOC docs.
- `e3a35b6` — fix UInt64 wrap-horizon year-math (~292M → ~584 years per Opus review). 1 LOC docs.

Both are docs-only, well within the polish envelope. nesquena's Mac build verification at `bec1ac8` still holds — no behavioural change since.

## Plan

If CI Test stays green on `e3a35b6`:

1. Squash-merge with `Co-authored-by: Claude Opus 4.7 (1M context)` (year-math fix attribution)
2. Push tag `v1.6.3` → `v1.6.4` (separate `git push` invocation)
3. CI builds + signs + notarizes the universal DMG → v1.6.4 GitHub Release
4. Update `desktop-app-mac.md` + `~/WebUI/AGENTS.md` + sync workspace
5. Verify `nesquena/hermes-webui#1733` auto-closes (cross-repo `Closes` reference)

PR #90, opened 2026-07-10
fix: allow artifact downloads in desktop client

maxon1233:
## Summary

Fixes desktop-client artifact downloads that currently do nothing when clicked.

- Route action-initiated HTML downloads through `WKDownload` before the generic `/api/` navigation guard.
- Attach the `WKDownloadDelegate` for action-initiated downloads as well as response-initiated downloads.
- Exempt the explicit artifact endpoints (`/api/media`, `/api/file/raw`, and `/api/folder/download`) from generic API-navigation cancellation so their `Content-Disposition: attachment` responses can reach the existing response-time download handler.
- Add regression coverage while preserving cancellation for ordinary API navigations.

## Reproduction

Hermes WebUI renders local artifact files as links such as:

```html
<a download href="api/media?path=...&download=1">📎 filename</a>
```

The Swift wrapper's navigation-action delegate canceled every `/api/*` request before WebKit could process the download action or receive the attachment response. The same authenticated `/api/media` endpoint returns `200` with `Content-Disposition: attachment` and downloads normally in Safari; the wrapper showed no visible result.

## Verification

- `git diff --check`
- Static policy-contract check confirming that `shouldPerformDownload` is evaluated before the API guard and that ordinary API navigation remains blocked.
- macOS runtime verification is intentionally left to CI/review: this patch requires an `NSSavePanel` for both a WebUI `📎` artifact link and a direct `/api/media?...&download=1` link. This Linux environment does not have Swift or Xcode installed.

### DIFF
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -1162,6 +1162,10 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         download.delegate = self
     }
 
+    func webView(_ webView: WKWebView, navigationAction: WKNavigationAction, didBecome download: WKDownload) {
+        download.delegate = self
+    }
+
     // MARK: - File upload
 
     func webView(
@@ -1193,6 +1197,15 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
             return
         }
 
+        // An HTML download attribute becomes a navigation action before a
+        // response is available. Handle it before the generic API guard below;
+        // Hermes WebUI artifact links use /api/media and would otherwise be
+        // cancelled before WKWebView can create a WKDownload.
+        if navigationAction.shouldPerformDownload {
+            decisionHandler(.download)
+            return
+        }
+
         // Defense-in-depth (#76): API endpoints should never become full-page
         // navigations. The WebUI's JS treats /api/* as fetch targets only, and
         // every API error response uses the JSON shape `{"error": "..."}` —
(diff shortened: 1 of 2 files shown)


nesquena (approved):
Verified on Mac agent at tip `a198aef`.

### Build/test results
- `swift build -c release` — Build complete, 0 warnings
- `swift test` — 32/32 passed (adds `testArtifactDownloadsAreNotCancelled`)
- Packaged via `bash build.sh` on `staging/v1.8.0` (this PR merged with #91): `.app` built, ad-hoc signed, ATS/Sparkle Info.plist keys verified intact

### Code review findings
- Traced the policy ordering in `webView(_:decidePolicyFor:)`: the `shouldPerformDownload` fast-path correctly runs before the `/api/*` guard, so `<a download href="api/media?...">` becomes a `WKDownload` instead of being cancelled. Ordinary API navigations still hit `.cancel`.
- The exemption list uses exact `url.path` matches — query strings aren't part of `path`, so `/api/media?path=...&download=1` matches while `/api/anything-else` stays blocked. Subpaths and trailing-slash variants still cancel, which keeps the #76 defense-in-depth intact.
- `webView(_:navigationAction:didBecome:)` was the missing delegate attachment for action-initiated downloads — without it the `WKDownload` would have no delegate and silently drop. Both response- and action-initiated paths now wire to the existing `WKDownloadDelegate` extension (save panel + error handling).
- Mirror-style test matches the suite's existing convention.

### Hardened on staging (non-blocking here)
- The fast-path honored `shouldPerformDownload` for **any** scheme, so a download-attributed `file://` link would have received `.download` instead of falling through to the scheme guard that cancels all `file://` navigation. Fixed on `staging/v1.8.0` (`5dd2af3`): the fast-path now skips `file://`; `blob:`/`data:` remain eligible since JS-generated exports rely on them.
- CHANGELOG entry (repo rule) added on staging in `291d03c`.

Merged into `staging/v1.8.0` together with #91 for the next release cut. Thanks @maxon1233 — tight, well-scoped fix with a clear repro.

PR #91, opened 2026-07-10
Strengthen the connection mechanism: fix tunnel bypass, support remote direct URLs, verify connections end-to-end

mreckhof:
Connection fixes:
- SSH mode loaded the configured Target URL directly, bypassing the ssh -L forward entirely (netstat showed a direct remote connection while the tunnel sat unused). AppDelegate.effectiveTargetURL() now resolves what browser windows load: in SSH mode always the tunnel entrance http://127.0.0.1:<local port>, in direct mode the Target URL. Applied to launch/reconnect, Cmd+N/Cmd+T, and Open in Browser.
- Direct mode rejected every non-localhost URL: the URLSession-based preflight, Test Connection, and health ping are subject to ATS (NSAllowsArbitraryLoads=false), which rejects plain-http to non-loopback hosts with -1022 before the (ATS-exempt) WKWebView gets a chance. All three now use ReachabilityProbe, a hand-rolled HTTP GET of hermes' /health over Network.framework (TLS for https), which ATS does not apply to.

Connection verification:
- The probe is tri-state — healthy (2xx + "status":"ok" body), reachable-but-unverified (port answers; e.g. reverse proxy up, hermes down), unreachable — surfaced in Test Connection and the direct-mode title dot (● / ◐ / ○). Dock badge only on unreachable.
- Test Connection in SSH mode runs a real non-interactive login (ssh -o BatchMode=yes user@host exit), distinguishing "no SSH" from "auth refused" (→ ssh-copy-id hint). After auth it verifies hermes end-to-end: through the live tunnel when every forward-shaping setting matches what it was built from, otherwise through a temporary ssh -N -L created from the field settings and torn down after the probe. Local port collisions report "port busy" (bind pre-check, so a stray listener can't masquerade as the tunnel); when our own live tunnel holds the port, the test detours through an ephemeral port. Editing any field or switching modes clears a stale test verdict.

Preferences UX:
- Mode segment is now "Tunnel: None | SSH" (persisted values unchanged); SSH sections dim/disable instead of hiding, so the fixed-frame layout keeps its shape. In SSH mode the Target URL row becomes a read-only derived tunnel-entrance label.
- Saving a plain-http Target URL to a non-loopback host asks a one-time per-host acknowledgment that credentials and session data cross the network unencrypted; loopback is exempt.

Hardening:
- SSH username/host fields reject option smuggling (leading "-", whitespace) at Save, Test Connection, and inside testAuth / testForward; BatchMode prevents invisible password prompts and a watchdog bounds stalled handshakes.
- The health probe builds its request from the percent-encoded URL path so encoded CR/LF cannot reshape the hand-rolled HTTP request.

60 unit tests (mirror-style, matching the existing suite) cover the effective-URL rule, probe endpoint/path derivation, health-response classification, auth-attempt classification, live-tunnel match rule, ssh identifier validation, and the plaintext-warning rule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **SSH mode no longer bypasses the tunnel** — the browser loaded the raw Target URL instead of the tunnel entrance, so the `ssh -L` forward came up but carried no traffic (`netstat` showed a direct connection to the remote host). `AppDelegate.effectiveTargetURL()` now returns `http://127.0.0.1:<local port>` in SSH mode and the Target URL in Direct mode, applied to launch/reconnect, new windows/tabs, and View → Open in Browser.
+- **Direct mode works with non-localhost URLs (Tailscale/LAN hosts)** — a Target URL like `http://my-box.ts.net:8787` failed with "Cannot connect" even when the server was up: the WKWebView is ATS-exempt (`NSAllowsArbitraryLoadsInWebContent`) but the URLSession preflight, Test Connection, and health ping were not, so ATS rejected their plain-http requests to non-loopback hosts with -1022. All three now use `ReachabilityProbe`, which speaks HTTP over Network.framework (not subject to ATS).
+
+### Added
+- **Tri-state `/health` verification** — the probe issues a real `GET /health` (hand-rolled over NWConnection, TLS for https) rather than just connecting, so it can tell a healthy hermes (2xx + `"status": "ok"`) from a reverse proxy answering for a dead backend (reachable) from nothing listening (unreachable). Surfaced in Test Connection and the direct-mode title dot (● / ◐ / ○); the Dock "!" badge fires only on unreachable.
+- **End-to-end SSH Test Connection** — verifies a real non-interactive login (`ssh -o BatchMode=yes user@host exit`), distinguishing "✗ No SSH" from "✗ Auth failed" (keys refused → `ssh-copy-id` hint). On success it confirms hermes end-to-end: through the live tunnel when every forward-shaping setting matches, otherwise via a temporary `ssh -N -L` built from the fields and torn down after probing — so "✓ Hermes OK" means auth + forward + verified hermes. Reports "✗ Port busy" when the local port is taken (bind pre-check) and clears a stale result when any field or the mode changes.
+- **Preferences: Tunnel selector** — the mode segment is now "Tunnel: None | SSH" (persisted values unchanged); the SSH sections dim/disable instead of hiding, and in SSH mode the Target URL row becomes a read-only derived tunnel-entrance label.
+- **Plaintext-HTTP acknowledgment** — saving a Direct `http://` URL to a non-loopback host prompts a one-time per-host warning that credentials and session data cross the network unencrypted. Loopback is exempt.
+
+### Security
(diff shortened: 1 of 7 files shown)


nesquena (approved):
Verified on Mac agent at tip `9146c4d`.

### Build/test results
- `swift build -c release` — Build complete, 0 warnings
- `swift test` — 60/60 passed (matches the PR's claim; six new mirror-style suites)
- Packaged via `bash build.sh` on `staging/v1.8.0` (this PR merged with #90): `.app` built, ad-hoc signed; `NSAppTransportSecurity`, `SUPublicEDKey`, `SUFeedURL` all verified intact in the built Info.plist

### Code review findings
- **Tunnel bypass fix confirmed real against main**: `startTunnel()` builds the `ssh -L` forward from the ssh fields but opened the browser at the raw `targetURL` — any non-localhost target rode straight past the tunnel. Traced `effectiveTargetURL()` to all four load sites (launch/reconnect, `reconnectInPlace`, Cmd+N/Cmd+T via `openNewBrowserSession`, View → Open in Browser). All correct.
- **`connectionMode` persistence preserved**: the SSH save path still writes `connectionMode`/`sshUser`/`sshHost`/ports; `targetURL` is intentionally left untouched so switching back to Direct restores the old value. Verified in the merged file, not just the diff.
- **`ReachabilityProbe` single-completion invariant holds**: state handler, receive loop, send completion, and the timeout all run on the same serial queue, and `finish()` is guarded — completion fires exactly once. The 256 KB receive cap bounds memory; the percent-encoded-path request build blocks CR/LF smuggling (covered by `testEncodedCRLFStaysEncodedInProbePath`).
- **Classifier validated against a live hermes**: `GET /health` on a running server returns `200` + `{"status": "ok", ...}` — exactly what `classifyHealthResponse` keys on, including surviving chunked encoding via substring match.
- **ssh hardening**: `isValidSSHIdentifier` blocks option smuggling (leading `-`, whitespace) at Save, Test Connection, and inside `testAuth`/`testForward`; `BatchMode=yes` prevents invisible password hangs and the watchdog bounds a stalled key exchange. `StrictHostKeyChecking=accept-new` matches the existing `start()` arguments, so the test exercises the same host-key path as the real tunnel.
- **Live-tunnel match rule**: the guard requires every forward-shaping field to match before the running tunnel may vouch for the settings, with the ephemeral-port detour when our own tunnel holds the configured port — the `LiveTunnelMatchTests` regression case (edited remote port must not report OK through the old forward) is the right invariant.
- Preferences: dim/disable relies on `isBezeled` to distinguish input fields from their labels inside `sshViews` — checked the `row()` helper; labels are `labelWithString:` (non-bezeled), fields are bezeled. Works. `modeChanged()` at the end of `buildUI` applies the initial state; no new rows, so no layout-descent risk.

### Non-blocking follow-ups
1. The hand-rolled request sends `Host: <host>` without the port — name-based reverse proxies on non-default ports technically expect `host:port`. Cosmetic until someone fronts hermes with strict vhost matching.
2. `testResultLabel` is 90pt wide; strings like "✗ Tunnel dead" may truncate — tooltips carry the detail, so fine for now.
3. CLAUDE.md still says `StrictHostKeyChecking=yes` while the codebase (pre-existing `start()` and this PR) uses `accept-new` — doc drift worth reconciling.

Merged into `staging/v1.8.0` together with #90 for the next release cut. Excellent work @mreckhof — the tri-state probe and the end-to-end Test Connection are exactly the diagnostics this app was missing.

PR #93, opened 2026-07-14
fix: swallow Escape when find bar is closed

roryford:
## Summary
- Escape key was propagating past the find bar when closed, causing unintended behavior downstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -1508,14 +1508,12 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         window?.makeFirstResponder(webView)
     }
 
-    // cancelOperation is sent by AppKit when the user presses Escape while
-    // the find field is first responder. Closing the bar here satisfies the
-    // CHANGELOG claim that Escape dismisses the bar.
+    // cancelOperation is sent by AppKit when the user presses Escape.
+    // Only act when the find bar is open — swallow it otherwise so Escape
+    // doesn't bubble up to NSWindow.performClose and hide the window.
     override func cancelOperation(_ sender: Any?) {
         if findBarVisible {
             hideFindBar()
-        } else {
-            super.cancelOperation(sender)
         }
     }
 


nesquena (approved):
Verified on Mac agent at tip `786706c`.

### Build/test results
- `swift build -c release` — clean at the PR tip (31/31 tests, v1.7.0-era base) **and** merged with current main (`release/v1.7.3`): clean, **62/62 tests pass**
- `bash build.sh` on the merged branch — `.app` built, ad-hoc signed; ATS/Sparkle Info.plist keys verified intact; smoke-launched against a live server

### Code review findings — the fix is more correct than the description claims
The branch is based on v1.7.0-era main, but the hunk merges cleanly with v1.7.2 (no overlap with the #90/#91 changes).

I reproduced the mechanism in a standalone AppKit+WKWebView harness (WKWebView as first responder, window controller overriding `cancelOperation`, Hermes-style `windowShouldClose` → `orderOut`):

1. **The old `super.cancelOperation(sender)` call was always a latent bug.** `cancelOperation(_:)` is one of the NSResponder key-binding methods that is *declared but not implemented* (`NSResponder.instancesRespond(to: #selector(cancelOperation))` is `false`). In the harness, the old path **throws `unrecognized selector` inside WebKit's `executeSavedCommandBySelector` IPC callout** on every bare Escape press. In the packaged app AppKit's event-loop exception guard swallows the throw (no crash report), leaving WebKit's command dispatch in an undefined state — which is what surfaces as the erratic window-hide (the last-window `windowShouldClose` → `orderOut(nil)` path at BrowserWindowController.swift:1289 makes the whole window vanish rather than just close).
2. **The new path is verified clean**: with the swallow behavior, the harness shows `cancelOperation` fires, no exception, `windowShouldClose` never fires, window stays visible — across repeated presses.
3. **No regression surface**: the `findBarVisible` branch (Escape closes the find bar, v1.6.1 behavior) is untouched, and any hypothetical window-level Escape default (e.g. full-screen exit via the responder chain) was already unreachable pre-PR — the old code intercepted and threw on the same path — so this change is strictly an improvement.

Merging via `release/v1.7.3` and cutting v1.7.3 with it. Thanks @roryford — small diff, real bug, and the comment you replaced accurately describes the new invariant.

PR #97, opened 2026-08-09
Fix long-conversation freeze on macOS (WKWebView)

itradesize:
Fixes the macOS app becoming permanently unresponsive when opening long conversations.

The issue is caused by the WebUI transcript virtualization scheduler repeatedly resetting its retry budget under WKWebView.

This adds a macOS-only compatibility guard that disables virtualization only when the affected scheduler behavior is detected. The server and WebUI deployment are unchanged.

Verified with focused regression tests and the full XCTest suite.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+- **Long conversations no longer enter WebKit's virtual-measurement retry loop** — the macOS client now installs a document-start compatibility guard for the upstream transcript-virtualization bug tracked by hermes-webui PR #6668. When the loaded scheduler still contains the faulty per-window retry reset, virtualization uses WebUI's supported non-virtualized fallback inside WKWebView only. Browser and iOS clients remain unchanged, and virtualization automatically resumes in the Mac client once the server supplies the corrected scheduler.
+
 ## [v1.7.3] — 2026-07-18
 
 ### Fixed
--- a/Sources/HermesAgent/BrowserWindowController.swift
+++ b/Sources/HermesAgent/BrowserWindowController.swift
@@ -77,6 +77,50 @@ private class TitleBarDragView: NSView {
 
 class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegate, WKNavigationDelegate, WKScriptMessageHandler {
 
+    /// WKWebView-only compatibility guard for the upstream transcript-virtualization
+    /// retry loop (hermes-webui PR #6668). The affected WebUI scheduler resets its
+    /// retry budget whenever WebKit's measured virtual window changes, so alternating
+    /// measurements can schedule layout work forever on long conversations.
+    ///
+    /// WebUI already provides `_virtualizeTranscript === false` as its supported
+    /// non-virtualized path. This document-start accessor remembers the server's
+    /// requested setting but reports it as disabled while the loaded scheduler still
+    /// contains the known bad reset. Once the upstream fix is deployed, the reset is
+    /// absent and the server setting takes effect automatically. Remove this shim once
+    /// supported WebUI versions universally include that fix.
+    static let longConversationWebKitCompatibilityScript = """
+        (function() {
+            const marker = '__hermesMacLongConversationCompatibilityPatch';
+            if (window[marker]) return;
+
+            const functionToString = Function.prototype.toString;
+            let requestedValue = window._virtualizeTranscript === true;
+
+            Object.defineProperty(window, marker, {
+                value: true,
+                configurable: false,
+                enumerable: false
+            });
+            Object.defineProperty(window, '_virtualizeTranscript', {
+                configurable: true,
+                enumerable: true,
+                get: function() {
+                    if (!requestedValue) return false;
+                    const scheduler = window._scheduleMessageVirtualMeasurementRefresh;
+                    if (typeof scheduler !== 'function') return false;
+                    const source = functionToString.call(scheduler);
+                    const compactSource = Array.from(source)
+                        .filter(function(character) { return character.charCodeAt(0) > 32; })
+                        .join('');
+                    return !compactSource.includes('_messageVirtualMeasurementRetryCount=0');
+                },
+                set: function(value) {
+                    requestedValue = value === true;
+                }
+            });
+        })();
+        """
+
     private var webView: HermesWebView!
     private var statusBar: NSView!
 
@@ -265,6 +309,17 @@ class BrowserWindowController: NSWindowController, NSWindowDelegate, WKUIDelegat
         prefs.setValue(true, forKey: "javaScriptCanAccessClipboard")
         prefs.setValue(true, forKey: "DOMPasteAllowed")
         config.preferences = prefs
+
+        // Install before any page content loads so the WebUI settings bootstrap
+        // cannot enable the affected virtualizer before this client-side guard.
+        // Main-frame-only keeps the override out of embedded documents.
+        let longConversationCompatibilityScript = WKUserScript(
+            source: Self.longConversationWebKitCompatibilityScript,
+            injectionTime: .atDocumentStart,
+            forMainFrameOnly: true
+        )
+        config.userContentController.addUserScript(longConversationCompatibilityScript)
+
         let pasteScript = WKUserScript(
             source:
                 "document.addEventListener('paste', function(e) { e.stopImmediatePropagation(); }, true);",
--- a/Tests/HermesAgentTests/LongConversationCompatibilityTests.swift
+++ b/Tests/HermesAgentTests/LongConversationCompatibilityTests.swift
@@ -1,0 +1,108 @@
+import JavaScriptCore
+import XCTest
+
+/// Regression coverage for the WKWebView-only transcript virtualization guard.
+///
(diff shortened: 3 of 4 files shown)


nesquena-hermes:
## Review — WKWebView long-conversation freeze shim (verified against live Core)

Reviewed this end-to-end. The diagnosis and the shim design are both correct, and I verified the Core-side contract it depends on directly against `nesquena/hermes-webui@master` rather than taking the PR's word for it.

**What the fix does:** installs a document-start, main-frame-only `WKUserScript` that intercepts `window._virtualizeTranscript`. It remembers the server's requested value but reports `false` (Core's supported non-virtualized path) *only while* the loaded scheduler still contains the buggy per-window retry-budget reset — detected by stringifying `window._scheduleMessageVirtualMeasurementRefresh` and checking for the `_messageVirtualMeasurementRetryCount=0` reset line. Once Core ships the fix, that substring is gone and virtualization resumes automatically.

**Verification I did against live Core (`master`):**
- **The referenced PR number is off.** The body/CHANGELOG cite `hermes-webui #6668`, but that isn't a merged core PR — the real fix is **#6654** (`fix(perf): don't reset virtual measurement retry budget on window-key oscillation`, commits `ec7f8bd`/`f1fb2ea`). Worth correcting the reference in the CHANGELOG + docs + the code comment so future maintainers can find it. (Non-blocking; the mechanism is right.)
- **The self-heal is precisely targeted — I confirmed it.** #6654's diff removes *exactly* the line `_messageVirtualMeasurementRetryCount=0;` from `_scheduleMessageVirtualMeasurementRefresh`, which is the exact substring your guard keys off. So once #6654 lands, `compactSource.includes('_messageVirtualMeasurementRetryCount=0')` goes false and the server setting takes effect — the shim self-removes as designed. ✅
- **The scheduler is reachable from the shim.** `ui.js` is loaded as a classic `defer` script (not an ES module), so `_scheduleMessageVirtualMeasurementRefresh` is a real `window` global that `Function.prototype.toString()` can read. ✅
- **No minification in Core's serving path**, so the whitespace-stripped substring match is stable against the served source (not a minified bundle). ✅
- **Correct today:** `master`'s scheduler still contains the reset at `static/ui.js:730`, so the shim correctly disables virtualization on current Core. And virtualization is **default-OFF** (`virtualize_transcript` is opt-in, `api/config.py`), so for the vast majority of users this is a no-op — it only engages for users who explicitly enabled it and are on an affected build.
- **Fail-closed** if the scheduler is missing (`typeof scheduler !== 'function' → return false`), and idempotent (`__hermesMacLongConversationCompatibilityPatch` marker).

**Tests:** the `LongConversationCompatibilityTests` are non-vacuous — they extract the *exact* production script from `BrowserWindowController.swift` source and run it in JavaScriptCore across all four states (buggy scheduler → disabled; fixed scheduler → server setting preserved; missing scheduler → fail-closed; disabled-server → stays disabled; idempotent double-install). Testing the real injected script rather than a hand-copied model is the right call.

**One design note (non-blocking):** matching on the stringified function body is inherently brittle — a Core refactor that keeps the bug but rewrites the reset (e.g. `_messageVirtualMeasurementRetryCount = 0` with spaces, or hoisting it into a helper) would make the guard read "fixed" and re-enable the buggy virtualizer. The whitespace-strip handles reformatting, but not a structural rewrite. Given virtualization is opt-in and the window until #6654 ships is short, this is an acceptable tradeoff — just worth a comment noting the guard is a temporary bridge tied to #6654's specific shape, and a reminder to delete the shim (you already have a "Removal" section — 👍) once #6654 is in every supported Core.

**Blocking item before merge:** this is a Swift change and Swift can't build on our Linux CI, so it needs a Mac-agent build + `swift test` pass, and an independent review posted from the `nesquena` account, before it merges (standard for this repo). @nesquena — could you run the Mac build/test loop on this branch? Once that's green with the reviewer approval, and the #6668→#6654 reference is corrected, I think this is good to ship.

nesquena-hermes:
## Correction + re-verification at `af0b04e10` — my earlier core references were wrong, and the self-heal guard has one real blind spot

Re-checked this against **live** core (`nesquena/hermes-webui@adbb3258`) rather than carrying my earlier review forward. The shim is still correct for users today, but I got two factual things wrong last time and owe you the correction.

### 1. My earlier core-PR references were wrong — and so was my correction

I told you the reference should be **#6654** rather than the #6668 in your CHANGELOG, and cited commits `ec7f8bd` / `f1fb2ea` as the merged fix. That was wrong:

| Claim I made | Actual live state |
|---|---|
| #6654 is the merged core fix | **#6654 is an open _issue_**, not a PR — `bug(perf): virtual measurement retry budget resets during window-key oscillation` |
| commit `ec7f8bd` is the fix | **No such commit exists** in core |
| commit `f1fb2ea` is on master | Exists (`f1fb2eacc`) but is **not** an ancestor of `origin/master` — it's branch history |
| #6668 is not a real core PR | Correct that it isn't merged — it is **CLOSED unmerged** |

The live vehicle is **open PR [#6717](https://github.com/nesquena/hermes-webui/pull/6717)** (`fix(perf): don't reset virtual measurement retry budget on window-key oscillation (#6654)`, head `05b6d5b47`, MERGEABLE, not yet merged). So the right reference for your CHANGELOG/docs/code comment is **issue #6654, fixed by PR #6717** — and the fix has **not** shipped yet. Sorry for sending you at a bad SHA; my earlier read used `git log --all`, which surfaced branch commits I then wrongly asserted were on master.

### 2. The shim is correct for every user today — verified, not assumed

`origin/master` still contains the buggy shape, so the guard correctly reports "core still broken" and keeps virtualization off:

```js
if(_messageVirtualMeasurementCycleKey!==cycleKey){
  _messageVirtualMeasurementCycleKey=cycleKey;
  _messageVirtualMeasurementRetryCount=0;   // ← the bug, still on master
}
```

Also still true, re-confirmed: `ui.js` is a classic `defer` script so the scheduler is a real `window` global; core's serving path isn't minified; and virtualization is opt-in (`virtualize_transcript`), so this is a no-op for nearly everyone.

### 3. New finding — the guard misclassifies a fix shape core actually wrote

I extracted the real scheduler source from every historical shape of this fix and ran **your exact guard predicate** (`BrowserWindowController.swift:112-115`, whitespace-stripped substring test) against each:

| Core scheduler shape | Actually fixed? | Your guard says fixed? | |
|---|---|---|---|
| `origin/master` today | no | no | ✅ |
| `d4e4ed3` — reset line deleted outright | yes | yes | ✅ |
| **`fc7eee3` — reset moved inside a once-per-burst guard** | **yes** | **no** | ❌ |
| `3a449dc` — seen-keys burst tracking | yes | yes | ✅ |
| **`05b6d5b` — #6717's current head** | yes | yes | ✅ |

The miss is `fc7eee3`, a real commit on #6717's own branch history. It fixes #6654 by moving the reset so it fires **once per measurement burst** instead of on every cycle-key change:

```js
if(!_messageVirtualMeasurementBurstActive){
  _messageVirtualMeasurementRetryCount=0;      // ← substring still present, but the bug is gone
  _messageVirtualMeasurementBurstActive=true;
}
```

Oscillation can no longer renew the budget, so the bug is genuinely fixed — but the literal `_messageVirtualMeasurementRetryCount=0` survives, so your guard reads it as "still buggy".

**This fails in the safe direction**, which is the important part: the shim keeps virtualization *disabled*, so nobody gets the freeze back. The cost is that the shim would never self-remove — Mac users would silently lose virtualization forever on a fixed core, with no signal. That's the opposite of the failure mode I warned about last time (I worried about re-enabling a buggy virtualizer; the realistic risk is actually a stuck-off shim).

**Good news:** #6717's *current* head is classified correctly, so if it lands as-is the shim self-removes exactly as you designed. The blind spot only bites if core lands the burst-guard variant instead.

**Suggested hardening** (cheap, and it makes the guard match intent rather than text): test for the reset *inside the cycle-key branch* rather than anywhere in the function — e.g. match `!==cycleKey` followed by the reset before the closing brace, or simply treat the presence of `_messageVirtualMeasurementBurstActive` / `_messageVirtualMeasurementSeenKeys` as a positive "core has the newer burst-tracking fix" signal. Either keeps the safe default while letting the shim actually retire.

### What I still can't do from here

Unchanged from my last review and still blocking: Swift can't build on our Linux CI, so this needs a Mac-agent `swift build` + `swift test` pass and an independent review posted from the `nesquena` account before it merges. The `LongConversationCompatibilityTests` look non-vacuous by reading (they extract the production script from source and exercise all four states in JavaScriptCore), but I have not executed them.

@nesquena — when you get a chance: Mac build/test loop on this branch, please.

### Summary

The diagnosis, the shim design, the fail-closed behaviour and the self-removal mechanism are all still right, and the implementation is careful. Three things to change before merge, none structural:

1. Reference **issue #6654 / PR #6717** in the CHANGELOG, docs and code comment — not #6668, and not the commits I gave you earlier.
2. Optionally harden the guard against the burst-guard shape above so the shim can actually retire.
3. Mac build/test + `nesquena` review.

_Gate: Codex (gpt-5.6-sol) route verified healthy this round; the substantive evidence here is my own harness replicating your guard predicate against five real core scheduler shapes extracted from core's git history, plus live state checks against `origin/master` and PR #6717. Reviewed at exact head `af0b04e10e1f6b611203ed3d4a514ee5362de5db`._

Issue #95, opened 2026-07-30
App Instantly freezes upon opening - tries to load a conversation


itradesize:
The macOS app instantly freezes upon opening. If I quickly switch to another conversation, it still works until I click on the most recent one, which causes it to freeze again - in a 'loading conversation..' state. 

This conversation loads fine in the browser or via the Hermex app.

What steps should I take to resolve this? What information do you need from me to help find a solution? 

<img width="1608" height="1015" alt="Image" src="https://github.com/user-attachments/assets/1dc0459e-5258-4710-a27c-175202f7769d" />

Andre-4711:
## Additional reproduction with concrete isolation and workaround

I hit what appears to be the same issue and was able to isolate it to the native app's WKWebView renderer.

### Environment

- macOS 15.7.7 (24G720), Apple Silicon / arm64
- Hermes Agent native macOS app 1.7.3
- Hermes Agent backend 0.20.0

### Trigger and measurements

- The affected conversation has 302 messages.
- The session API response is approximately 5.82 MB.
- One historical tool/message payload accounts for approximately 5.38 MB.
- The same conversation loads successfully in Chromium.

When the native app opens that conversation:

- the UI remains at `Loading conversation...` indefinitely;
- the Swift host process remains close to 0% CPU;
- `com.apple.WebKit.WebContent` stays at approximately 100% CPU for minutes;
- the backend remains healthy (`/health` reports `status: ok`);
- there are no active runs or streams;
- the exact session API request returns HTTP 200 in approximately 0.23 seconds.

This strongly suggests a WKWebView rendering loop on the multi-megabyte historical payload rather than a backend, provider, or persistence failure.

### Why restarting does not help

The app restores the last selected conversation from the WebKit LocalStorage key `hermes-webui-session`. Relaunching therefore immediately loads the same pathological conversation and the newly spawned WebContent process returns to 100% CPU.

### Minimal recovery that worked

1. Quit only the native app.
2. Back up its `localstorage.sqlite3` database.
3. Remove only the `hermes-webui-session` row when it points to the affected conversation.
4. Relaunch the app.

After that recovery:

- the conversation itself and all messages remain intact;
- the native app opens normally;
- WebContent returns to 0% CPU;
- ordinary navigation within the app works again;
- the backend remains healthy.

Reopening the same oversized conversation in the native app is expected to reproduce the hang; Chromium is a viable temporary access path.

### Suggested fix directions

[...]

PR #99, opened 2026-08-14
Fix WKWebView stuck on loading conversation

keithah:
WKWebView can leave the WebUI on its loading placeholder when a post-response navigation fails or the WebKit content process terminates.

Changes:
- handle both provisional and committed navigation failures
- reload in place after WKWebView content-process termination
- inject a native-shell marker so hermes-webui can use a smaller first transcript window
- add a macOS-14 PR workflow that runs tests and uploads an installable app artifact

Verification: git diff --check and JavaScript syntax check pass locally. Full Swift build/tests run in macOS CI.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -1,0 +1,39 @@
+name: macOS PR Build
+
+on:
+  pull_request:
+    branches: [main, master]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: macos-pr-build-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Resolve Swift packages
+        run: swift package resolve
+
+      - name: Run unit tests
+        run: swift test
+
+      - name: Build release app
+        run: bash build.sh
+
+      - name: Archive installable app
+        run: ditto -c -k --sequesterRsrc --keepParent "Hermes Agent.app" "Hermes-Agent-pr-${{ github.event.pull_request.number || github.run_number }}.zip"
+
+      - name: Upload app artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: Hermes-Agent-pr-${{ github.event.pull_request.number || github.run_number }}
+          path: Hermes-Agent-pr-${{ github.event.pull_request.number || github.run_number }}.zip
+          if-no-files-found: error
+          retention-days: 14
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## Review — the WKWebView fixes were reverted; only the CI workflow remains

Thanks @keithah. I reviewed this at head `faba31b` against `main`, and there's a mismatch between the PR's stated purpose and what it now delivers that I want to flag before this goes further.

### What the branch actually changes vs. main

The net diff against `main` is a **single file** — `.github/workflows/pr-build.yml` (+39 lines). `Sources/HermesAgent/BrowserWindowController.swift` is **byte-identical to main**: the four WKWebView commits (`d818dea` recover-on-load-failure, `f27abd8` replay-cache recovery, `7a1a748` bound oversized fields, `d818dea`/`7a1a748` shims) were all removed by the final commit `faba31b` "rely on WebUI transcript rendering bounds" (−72 lines). So despite the title "Fix WKWebView stuck on loading conversation," **no WKWebView fix ships here** — the branch is now a CI-workflow-only change.

That's a legitimate engineering conclusion (if the real fix belongs in WebUI Core's transcript rendering rather than the Mac shell), but it means the PR as-titled doesn't address the bug, and #95 ("App instantly freezes… tries to load a conversation") stays open.

### On the remaining CI workflow (`pr-build.yml`)

Two things to reconcile before it's mergeable on its own merits:

1. **The `swift test` step duplicates the existing `test.yml`.** `main` already runs `swift test` on every PR via `test.yml` (`Test` workflow, `pull_request: [main]`). `pr-build.yml` re-runs it, so PRs would run the unit suite twice. If the intent is a PR *build*+*artifact* (not tests), drop the `Run unit tests` step and let `test.yml` own testing; keep only `swift package resolve` → `bash build.sh` → artifact upload.
2. **The value it does add is real:** `build-release.yml` only builds on `v*` tags, so there's currently no PR-time verification that `build.sh` still produces a working `.app`. A build+installable-artifact job on PRs is worth having (`build.sh` uses ad-hoc signing, `codesign --sign -`, so it needs no secrets in CI — good). If you want to land that, please split it into its own PR titled for what it is (e.g. "ci: build installable macOS PR artifact") so it's reviewable as an infra change, not tangled with a WKWebView fix that isn't here.

### Suggested path

- If you still intend to fix #95: restore the WKWebView recovery work (or the Core-side transcript-bounds change you concluded was the right layer) and re-point the PR at it, with a note on where the fix landed. Happy to help verify against a live Core once there's Swift to review — this one needs a real Mac `swift build`/`swift test` cycle, so I'd want @nesquena to run that.
- If the WKWebView fix is genuinely deferred to Core: I'd **close this PR** and open a focused `ci:` PR for the PR-build workflow (minus the duplicate test step), so #95 stays clearly open and the CI change lands on its own.

Not merging — flagging the title/content mismatch and the duplicate-test-step so you can decide which way to take it. What's your intent for the WKWebView side?

*(Reviewed at head `faba31b`; net diff = `pr-build.yml` only; `BrowserWindowController.swift` identical to `main`. No merge/label action taken.)*

keithah:
Realized we should fix this all from the WebUI part, and not the Swift app. The WebUI fix is in https://github.com/nesquena/hermes-webui/pull/7040.

